PR #4, opened 2024-06-16
changedtimestamps

brandesdavid:
## Describe the issue:

when the bot was offline, and messages were still pending to be processed, the bot instantly added all recognized stations to the database with the same timestamp. 

added check for null types of info dict variable (no overflowing error logs),
added timestamp to the go backend struct,
get timestamp from message (which is rounded and cleaned, no traceback) to the message handler and process message in bot package.

when adding new entries, which have been pending, their respective timestamps are placed.


packages/FreiFahren_BE-NLP/db_utils.py
```diff
@@ -53,17 +53,20 @@ def insert_ticket_info(
 
     # Prepare the JSON data payload
     data = {
+        'timestamp': timestamp.isoformat(),
         'line': line,
         'station': station_name,
         'direction': direction_name,
         'author': None,
-        'message': message
+        'message': 'from bot'
```
johan-t: Can you Ignore this message in getRecentTicketInspectors so that we can use this an identifier for messages that are coming from the bot and messages coming from the user but still avoid displaying it to the user.

packages/FreiFahren_BE-NLP/main.py
```diff
@@ -145,8 +103,12 @@ def process_new_message(author_id, message, current_time, conversations):
...
-        current_time = datetime.datetime.now()
-        
-        process_new_message(author_id, message, current_time, conversations)
+        timestamp = datetime.fromtimestamp(message.date, utc)
+        if timestamp.second or timestamp.microsecond:
+            timestamp = timestamp.replace(second=0, microsecond=0) + timedelta(minutes=1)
+        else:
+            timestamp = timestamp.replace(second=0, microsecond=0)
+            
+        process_new_message(timestamp, author_id, message, conversations)
```
johan-t: The function takes two parameters but 4 are given

johan-t: I fixed it

johan-t (changes requested):
Good changes small fixes and it can be merged

johan-t:
Also one more thing: I noticed that in the function CheckIfModifiedSince the lastModified that is being returned is still in CET this causes caches to be invalid so this needs to be fixed before we can merge

johan-t:
> Also one more thing: I noticed that in the function CheckIfModifiedSince the lastModified that is being returned is still in CET this causes caches to be invalid so this needs to be fixed before we can merge

The timestamp is also in the wrong format

johan-t:
I think it would also be beneficial if you could write a sentence or two either in the readme or in a docs document on the format of the timestamp so mention that everything is in utc and the format and explain the decision behind it.

johan-t:
> Also one more thing: I noticed that in the function CheckIfModifiedSince the lastModified that is being returned is still in CET this causes caches to be invalid so this needs to be fixed before we can merge

You can see this for example if you call the /getSegmentColors endpoint the lastModified time will be in CET

johan-t:
Also I just noticed the lastReportTime in the localstorage is still in a different format, could you fix this?

brandesdavid:
> lastReportTime

fixed the ModTime to be changed to UTC

brandesdavid:
> Also I just noticed the lastReportTime in the localstorage is still in a different format, could you fix this?

do you mean something like 1718712179066? they are the milliseconds starting at 1970 and in UTC, converting it makes it easier as we don't need to handle strings like YY:MM:DD:SS

johan-t:
> > Also I just noticed the lastReportTime in the localstorage is still in a different format, could you fix this?
> 
> 
> 
> do you mean something like 1718712179066? they are the milliseconds starting at 1970 and in UTC, converting it makes it easier as we don't need to handle strings like YY:MM:DD:SS

I feel like it is better to have it in the one format that we use everywhere especially since we are not comparing with the timestamp that often meaning we don't have any negative performance implications, but it is easier to read (easier to notice bugs) and more consistent

brandesdavid:
> > > Also I just noticed the lastReportTime in the localstorage is still in a different format, could you fix this?
> > 
> > 
> > do you mean something like 1718712179066? they are the milliseconds starting at 1970 and in UTC, converting it makes it easier as we don't need to handle strings like YY:MM:DD:SS
> 
> I feel like it is better to have it in the one format that we use everywhere especially since we are not comparing with the timestamp that often meaning we don't have any negative performance implications, but it is easier to read (easier to notice bugs) and more consistent

changed the lastreporttime to the iso8601 string, which gets calculated back to milliseconds to check for 15 minute reportin span

brandesdavid:
@johan-t

johan-t:
@brandesdavid good work! Can be merged imo

johan-t (approved):

PR #13, opened 2024-06-22
fix: risk prediction data is always up to date

johan-t:
## Describe the issue:
1. The preloaded risk data was not refreshed when a new inspector was added causing a shift after the newest risk data was requested
2. the risk data was always one minute behind the real time

1. I used React Context to create a global risk data state as it is being updated in multiple places. The context is updated when we are loading the app for the first time, when a new inspector is added, every 30 seconds that the risk line layer is loaded
2. For some reason one minute was added to the timestamps of the bot and in the backend when reporting a new station this caused the database data to be in the future and the risk model which is getting the current data, always had to wait for a minute until it used to data from json.


brandesdavid (approved):
lgtm

PR #15, opened 2024-06-25
Johan/fre 113 log files

### LINKED ISSUE #1: Remove Aussteigen logic

johan-t:
## Describe the issue:
Hard to debug in prod when we had little info about what happend

- I created a logger set up where we use the python native logging package to log.
- I logger.info the general important stuff and when saving user data (as it rounds the timestamp)
- logger.debug for all of the function calls
- logger.error for fatal errors

- fixes: #1


packages/FreiFahren_BE-NLP/main.py
```diff
@@ -45,9 +46,6 @@ def extract_ticket_inspector_info(unformatted_text):
 
     # With the found info we can cross check the direction and line
     if found_line or found_station or found_direction:
-        # direction and line should be None if the ticket inspector got off the train
-        handle_get_off(text, ticket_inspector)
```
brandesdavid, line 49: why did you remove this?

packages/FreiFahren_BE-NLP/main.py
```diff
@@ -81,10 +79,12 @@ def process_new_message(timestamp, message):
                 info.get('direction'),
             )
     else:
-        print('No valuable information found')
+        logger.info('No info found')
```
brandesdavid: Please make this more precise, like where this Log has been made and what exact information has not been found

packages/FreiFahren_BE-NLP/process_message.py
```diff
@@ -2,21 +2,34 @@
...
 from fuzzywuzzy import process
 from NER.TransportInformationRecognizer import TextProcessor
+from logging_utils import setup_logger
 import os
 
+logger = setup_logger()
+
 
 def load_data(filename):
+    logger.debug('loading data')
```
brandesdavid: Which data are we loading?

packages/FreiFahren_BE-NLP/process_message.py
```diff
@@ -46,6 +61,8 @@ def process_matches(matches_per_word):
 
 
 def find_line(text, lines):
+    logger.debug('finding the line')
+
     formatted_text = format_text_for_line_search(text)
```
brandesdavid, line 66: Maybe add these to sub categories,

like  logging

"Finding line"
...
"Finding line(2/3): formatting text"
...
"Finding line (2/3): couldn't format text error"

brandesdavid (changes requested):
logger looks good.
about the logging, i think you should add more context (also more errors), so the log files have better traceability and better reading

like e.g.

"(utils.py) searching for lines"
"(Utils.py) searching for lines (3/4): get line direction..."
"(Utils.py) searching for lines (3/4): error at get_line_direction: "\n" not found in database"

Etc...

packages/FreiFahren_BE-NLP/main.py
johan-t: Check the commit message

packages/FreiFahren_BE-NLP/process_message.py
johan-t: I cannot log the errors as they are unexpected everytime that an error is expected we are logging the exception to the file to make it traceable but by keeping the console clean of logs we can leave the console for errors so checking errors still has to be done in the console

johan-t: @brandesdavid I don't really understand the sub categories could you explain?

johan-t: > I cannot log the errors as they are unexpected everytime that an error is expected we are logging the exception to the file to make it traceable but by keeping the console clean of logs we can leave the console for errors so checking errors still has to be done in the console

We can also build an automatic tool that creates a github issue for each error appearing in the console

johan-t:
Also @brandesdavid the logging timestamps are in utc not sure if it is a good idea to do this or if CET is better, what do you think?

PR #19, opened 2024-06-29
Watcher

johan-t:


johan-t:
So I think the watcher should work like this:

Telegram Bot:

Watch the console if there is anything printed into the console. If this is the case call the do_healthcheck function with the url of the telegram healtheck endpoint to check if the entire server is down. Then send the report to the user with information about the error message that it noticed and if the server is continuing to run or if it is completely down.
Backend

Once an error occurs in the backend it will make a request to the /failure-report endpoint in the watcher to notify it about an error occuring.
It will then proceed to call the test_backend_apis function that calls each api and checks if they are responding and if the format is as expected (for example / recent should always return at least one inspector)
After calling each useable endpoint it will call the do_healthcheck function and call the backend healthcheck endpoint to check if the entire server is down.
It will then send the results of the down time analysis to the user in the chat for instance: "X error occured, y out of z endpoints responded as expected, these endpoints responded not as expected: ..."
It should continue to run the test_backend_apis function every minute or so to make sure it is running even when there was no error

johan-t:
Also when the user runs /healthcheck in the chat it should do run the script again for bot and backend

johan-t:
also no abbreviations such as errh use error_http etc.

johan-t:
I think the do health check functions where we are requesting the /healthcheck endpoint of bot and backend can be used as a single function to make the code easier to maintain.

PR #35, opened 2024-07-03
Johan/fre 222 avoid instantly asking for users location

johan-t:
## Describe the issue:
- When I was in subway it often happend that I was asked for my location even when I have shared it, due to a small outage in the network.
- I have noticed that users who dont share their location always click the modal down because it is being opened instantly which does not give them time to first process the new info on the map. 

I think this will increase the number of people submitting their location using the form as we avoid overwhelming the user with an instant request.


brandesdavid (approved):
LGTM

PR #48, opened 2024-07-07
Fix header date format

mclrc:
The date format we use for the `If-Modified-Since` header is non-standard. This leads to issues on android because the network stack can add it's own caching headers with the specced format, which will result in a code 500 with current parsing logic.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since

### DIFF
--- a/packages/backend/utils_test/CheckIfModifiedSince_test.go
+++ b/packages/backend/utils_test/CheckIfModifiedSince_test.go
@@ -26,14 +26,14 @@ func TestCheckIfModifiedSince(t *testing.T) {
 		},
 		{
 			name:             "Correct header, not modified",
-			ifModifiedSince:  "2024-01-01T12:00:00Z",
+			ifModifiedSince:  "Mon, 01 Jan 2024 12:00:00 GMT",
 			lastModified:     lastModifiedTime,
 			expectedModified: false,
 			expectError:      false,
 		},
 		{
 			name:             "Correct header, modified",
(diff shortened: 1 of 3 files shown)


packages/backend/utils_test/CheckIfModifiedSince_test.go
```diff
@@ -27,14 +26,14 @@ func TestCheckIfModifiedSince(t *testing.T) {
 		},
 		{
 			name:             "Correct header, not modified",
-			ifModifiedSince:  "2024-01-01T12:00:00Z",
+			ifModifiedSince:  "Mon, 01 Jan 2024 12:00:00 GMT",
```
johan-t, line 29: Should we be using GMT here or UTC?

johan-t (approved):
Good catch. Before merging please make sure that the caching still works, just in case as it is not being tested automatically.

packages/backend/utils_test/CheckIfModifiedSince_test.go
mclrc: very confusing but GMT at the end usually means UTC nowadays. check the output of `new Date().toUTCString()`.

johan-t: > very confusing but GMT at the end usually means UTC nowadays. check the output of `new Date().toUTCString()`.

lol

PR #62, opened 2024-07-23
refactor: Inspectors endpoint uses rest api best practices

johan-t:
## Describe the issue:
The /newInspectors and /recent endpoints were not built using REST API best practices this caused us needing more endpoints than necessary. For example /recent, /newInspector and /stats can be handled from one endpoint which is /Inspectors depending on the meta data and parameters it will be handled or these use cases.

Also we were using Station names instead of ids this is a bit dangerous considering how we had inconsistency issues.

I have merged the /recent and /newInspector endpoints to GET and POST Inspectors Endpoint. As an input for POST Inspectors we are now using Ids instead of names, note that this requires changes in the mobile app @mclrc 
I have also changed GET Inspectors so that you can now add an optional start and endtime and it will return the reports of that time, this will allow to use the list functionality for all of the reports in the last 24 hours. @mclrc could you update this functionality in the mobile app?

/stats is also deleted and now we are filtering for the past 24 hours and just getting the length of the returned array.

These changes require some small updates to the app but I think it will make the codebase easier to maintain as we have less code and more functionality.


packages/backend/api/getSegmentColors/getSegmentColors.go
```diff
@@ -30,9 +30,9 @@ import (
 // @Success 304 {none} nil "No changes: The data has not been modified since the last request date provided in the 'If-Modified-Since' header."
 // @Failure 500 "Internal Server Error: Error during the processing of the request."
 //
-// @Router /risk-prediction/getSegmentColors [get]
+// @Router /risk-prediction/segmentColors [get]
 func GetSegmentColors(c echo.Context) error {
-	logger.Log.Info().Msg("GET /risk-prediction/getSegmentColors")
+	logger.Log.Info().Msg("GET /risk-prediction/segmentColors")
```
mclrc: should we align our url casing while we're at it? segment-colors?

johan-t: yeah good point, i will fix this

PR #64, opened 2024-08-04
Johan/fre 250 watcher does not watch nlp bot console

johan-t:
## Describe the issue:
- the watcher did not watch the issues in the console of the sub threads in real time, only after being shut down it would report all of the issues
- the architecture of the bots was a bit byzantine which made the code harder to maintain or to add something on top of it. This was due to the multiple threads being run in different places. 

I centralized the architecture, we now have a main thread that runs the server, and starts the different sub processes as threads in a single location using a try except, if anything is expected it will be reported in the chat and logged, that way we can catch all of the errors in real time and the multi threaded development is a lot easier as there is a single point of failure.

I also refactored a lot and reused a lot of code snippets that way the code will be easier to maintain as we have less code. I tried to develop using the KISS principle.

Just in case I also added better documentation for project and how to start development.

- we would need to change the .env on the server
- I also changed the dockerfile for the new set up system
- I am not sure if there are additional things that we need to change on the server besides the .env, could you check @brandesdavid?


brandesdavid (approved):
LGTM

johan-t:
> LGTM

@brandesdavid but do we need to change something else besides the .env? for example something related to the production flask server?

PR #156, opened 2024-10-03
Hate speech filter

brandesdavid:
## Describe the issue:

We need a hate speech filter 

Add transformers library and run it locally


johan-t (approved):
llllllgtm

PR #241, opened 2024-11-15
Prepare for App Store release

mclrc:


packages/app/assets/locales/de.json
johan-t, line 1: is there any reason to not have the legal disclaimer of the mobile and web app in sync?

packages/app/src/components/Disclaimer/index.tsx
```diff
@@ -4,11 +4,12 @@ import { DateTime, Duration } from 'luxon'
 import { Box, Text } from 'native-base'
 import { useEffect, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
-import { View } from 'react-native'
+import { View, Linking } from 'react-native'
```
johan-t: lets merge with all of the linting erros fixed

packages/frontend/src/pages/Support/Support.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className="legal-text">
+      <h1>Support</h1>
+      <br />
+      <h2>iOS/Android</h2>
+      <p>Bei Problemen oder Anregungen bezüglich der Mobile-App, melde dich bitte bei Moritz unter <a href="mailto:moritzamando@proton.me">moritzamando@proton.me</a></p>
+      <h2 style={{ marginTop: 16 }}>Web</h2>
+      <p>Falls du ein Problem mit der Website hast, kontaktiere bitte Johan unter <a href="mailto@johan@trieloff.net">johan@trieloff.net</a></p>
+      <div>
+        <ul style={{ display: "flex", flexDirection: "row", gap: 10 }}>
```
johan-t: please avoid using inline css as it is harder to debug and nicer to separate concerns

packages/frontend/src/pages/Support/Support.tsx
```diff
@@ -0,0 +1,26 @@
+import { Link } from "react-router-dom"
+
+const Support = () => {
+  return (
+    <div className="legal-text">
```
johan-t: Can you add english support to this like it is done with the other strings using the language lib?

packages/app/assets/locales/de.json
johan-t: Imo we dont have to keep everything in sync but the legal stuff should be in sync if possible because this is the part of the app where if we fuck up it has the highest consequences and if there are two legal versions for everything there is a higher chance of fucking something up

johan-t:
Also for the future imo it is always cool to write a good pr description with some screenshots it makes it easier to debug issues in the future using git blame if we merge because it is easier to understand why something was introduced 6 months down the line

packages/app/assets/locales/de.json
mclrc: the reason is that the web legal disclaimer is pretty specific to tickets which we need to avoid in the app. would be nice to keep in sync eventually but not for initial reviews

johan-t: Okay I get that

johan-t:
Also do you think that after the review has passed, it would be possible to add the mentions of ticket inspectors again as I think this would help with SEO in the future?

mclrc:
probably, although i don't know if the placeholders will really be picked up by crawlers anyway, especially since they are js rendered

johan-t (approved):
🤩

johan-t:
> probably, although i don't know if the placeholders will really be picked up by crawlers anyway, especially since they are js rendered

I mean at least in the web app it should be fine, right?

mclrc:
> I mean at least in the web app it should be fine, right?
js rendered content being picked up by crawlers is kind of hit or miss, we'd need SSR to be sure. but yeah we should be able to revert the copy changes once the review is through

johan-t:
> > I mean at least in the web app it should be fine, right?
> > js rendered content being picked up by crawlers is kind of hit or miss, we'd need SSR to be sure. but yeah we should be able to revert the copy changes once the review is through

I agree I also think that SEO is not important right now, we can build a proper landing page if we want to be good at SEO

PR #242, opened 2024-11-16
feat: support trams

johan-t:
### Issue Summary:
The app currently supports U-Bahn and S-Bahn stations but lacks support for trams. This has been a long-standing request from users, some of whom have started bypassing this limitation by reporting tram stations through the description field.

### Solution Details:
To address this issue, the following changes were made:
1. **Expanded Station and Line Lists**:
   - Added all tram stations to the `StationsList.json`.
   - Updated `LinesList.json` to include all tram lines.
2. **Migration for Updated Station IDs**:
   - Implemented a migration script to update station IDs and their prefixes. For example, `SU-A` (Alexanderplatz S-Bahn) is now `SUM-A` to account for the tram station at Alexanderplatz.
3. **Report Form Update**:
   - Ensured tram stations are now included in the report form.
4. **Rewrote Segment Handling**:
   - Converted the R script by @jfsalzmann into a Python script (`segments.py`). This eliminates the team's dependency on @jfsalzmann for modifications and ensures the script can be maintained more flexibly.
   - Replaced the old segments with new ones that include tram data.
   - Moved segment data for the frontend to the public folder to prevent unnecessary increases in the bundle size.
5. **Telegram Bot Adjustments**:
   - Removed the filter that previously excluded trams from being recognized by the bot.
6. **Synonyms for Tram Stations**:
   - Added synonyms to ensure consistent and accurate identification of tram stations.
  
Bildschirmfoto 2024-11-16 um 13 27 41

### Additional Notes:
- Support is limited to metro trams, as these are the only ones currently subject to checks.
- Due to the migration there will be a small downtime when this is merged


johan-t:
@mclrc This requires the Files in the App to also be changed. It would be ideal if you find a way to request the station and lines data from the backend in the first use and then cache it locally to avoid having a new request every time. That way we can make it easier to change things in the future. But for now it is also fine to just swap the files for the new ones.

PR #246, opened 2024-11-17
refactor: better sharing

brandesdavid:
include station, line and direction
add a check for empty string in marker modal for line and direction

### DIFF
--- a/packages/frontend/public/locales/de.json
+++ b/packages/frontend/public/locales/de.json
@@ -71,8 +71,8 @@
         "disclaimer": "Daten vielleicht ungenau."
     },
     "Share": {
-        "title": "Schau dir diese App an",
-        "text": "Siehe wo Kontrolleure in Berlin sind",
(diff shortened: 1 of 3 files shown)


johan-t (approved):
This looks pretty good, if you want to perfect it you could add dynamic links so that when a person clicks on the link it will open the correct inspector. But I am not going to block it because of this. Imo we can add this later this is already a great improvement.

johan-t:
Also one more thing, please update the description so that it is easier to know without the necessary context what this is about. This will help with debugging. Also maybe include the problem this solves so that down the line we will know why this was important.

johan-t:
This is going to play nicely with https://linear.app/freifahren/issue/FRE-369/share-button-inside-of-marker-modal-should-be-bigger where this sharing button will be reworked as a whole.

johan-t:
@brandesdavid can this be merged?

PR #266, opened 2024-11-18
App: EAS-independent fastlane build setup

mclrc:
## Describe the issue:
Apps using Expo usually use their EAS build tool, primarily a CLI for their hosted cloud infrastructure, that simplifies the building/signing/submitting/updating process. It tool does kind of support local builds but is mainly designed to push you towards the paid service. It also makes certain kinds of configuration more difficult.
This PR gets rid of this reliance and sets up a 100% local code-to-store build setup based on fastlane instead of EAS while still using Expo to simplify native config
This setup isn't very straightforward and apple especially has quite involved code signing and certificate requirements. Especially as long as we don't have something like Bitwarden to safely share secrets I don't think anyone else should set up their system for production builds. Local builds/testing works the same as before.


johan-t:
@mclrc good description

johan-t:
@mclrc If you think that it is a priority I can setup 1password using the github student developer pack, we get it for free for 1 year I think

johan-t (approved):
tbh I dont really have enough knowledge about mobile development to add anything to this conversation, but the changes seem to be good 👍

mclrc:
I actually think it's fine, I can handle app releases for now and ideally we should probably set up a CI server with credentials instead of having keys lying around on personal laptops anyway

PR #272, opened 2024-11-19
App: Add tram support, fetch lines/stations from backend

mclrc:
## Describe the issue:
app didn't have tram support (probably lead to some crashes) and used duplicaed lines/station data
add tram support + fetch stations/lines from the backend


packages/app/.prettierrc.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  ...require("../../.prettierrc.js")
```
johan-t, line 2: Wouldnt it be better to use a global prettier config so that all of the projects are unified?

johan-t (commented):
btw, I think for the time being you dont have to wait for an approval until something is merged if you are just changing something on the app that is not very controversial. I am not able to give a code review with react native stuff. But having a nice Pr description is still nice to keep track over the changes. Maybe even screenshots if you like to

packages/app/.prettierrc.js
mclrc: yes but that quickly gets messy, for editor plugins/pre-commit-hooks/etc it really makes life a lot easier if you have a config in your cwd

PR #274, opened 2024-11-19
Add tracking to app

mclrc:
## Describe the issue:
We didn't have tracking in the app
Added a "website" in Pirsch and implemented events into the app
I had to use raw HTTP(S) as the Pirsch SDK does not support Hermes/React Native. Shouldn't be a problem though.

### DIFF
--- a/packages/app/src/Main.tsx
+++ b/packages/app/src/Main.tsx
@@ -1,11 +1,20 @@
+import { noop } from 'lodash'
 import { View } from 'native-base'
+import { useEffect } from 'react'
 
 import { FFMapView } from './components/FFMapView'
 import { UIOverlay } from './components/UIOverlay'
+import { client } from './tracking'
 
-export const Main = () => (
-    <View width="100%" height="100%">
-        <FFMapView />
-        <UIOverlay />
-    </View>
-)
+export const Main = () => {
+    useEffect(() => {
+        client.hit().catch(noop)
+    }, [])
+
+    return (
+        <View width="100%" height="100%">
+            <FFMapView />
+            <UIOverlay />
+        </View>
+    )
+}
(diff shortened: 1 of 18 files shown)


packages/app/src/Main.tsx
```diff
@@ -1,14 +1,13 @@
...
 
 import { FFMapView } from './components/FFMapView'
 import { UIOverlay } from './components/UIOverlay'
-import { client } from './tracking'
+import { track } from './tracking'
 
 export const Main = () => {
     useEffect(() => {
-        client.hit().catch(noop)
+        track({ name: 'app-opened' })
```
johan-t: I think instead of sending this as an event you can send it as a `/hit` request and attach some info about the device that way it is closer to the real sdk and you will have a better dashboard

johan-t: https://docs.pirsch.io/api-sdks/api-guide

mclrc: good point

packages/app/src/components/UIOverlay/Report/ReportSheet.tsx
```diff
@@ -89,7 +89,7 @@ export const ReportSheet = forwardRef((_props: PropsWithChildren<{}>, ref: Ref<R
     const onSubmit = async () => {
         if (!isValid) return
 
-        track({ name: 'report-submitted', timeTaken: Date.now() - openedAt.current! })
+        track({ name: 'Report Submitted', timeTaken: Date.now() - openedAt.current! })
```
johan-t: I think instead of using timeTaken you can use the Pirsch native "duration" param but you will have to check on that

johan-t (approved):

PR #279, opened 2024-11-20
Add sentry to app

mclrc:


packages/app/App.tsx
```diff
@@ -14,10 +15,16 @@ import { GestureHandlerRootView } from 'react-native-gesture-handler'
...
 
 import { queryClient } from './src/api'
+import { config } from './src/config'
 import i18n from './src/i18n'
 import { Main } from './src/Main'
 import { theme } from './src/theme'
 
+init({
+    dsn: config.SENTRY_DSN,
+    debug: true, // If `true`, Sentry will try to print out useful debugging information if something goes wrong with sending the event. Set it to `false` in production
```
johan-t: Let's set this to false and toggle true when we are developing locally

PR #283, opened 2024-11-20
FRE-392 FRE-377 FRE-387 FRE-382 FRE-374 App polish

mclrc:
image
image


johan-t:
Can you attach screenshots how the app looks like after the changes?

johan-t:
Looks great 👍

PR #287, opened 2024-11-21
Refactor theming with Restyle + design experimentation

mclrc:
## Describe the issue:
Previously, the app was using `native-base` for styling/themes. Besides the fact that it is now deprecated, it had some dependency issues and akward quirks that made using it for a proper design system a bit of a pain.
Switched to `@shopify/restyle`, which provides much more ergonomic systems, and implemented unified themeing in the app
PR currently also contains a new colorscheme, however this is mostly for experimental purposes and does not have to be released. Changing it should be very easy with restyle.


johan-t (approved):
I think this is okay for now. But before settling on a final design we should test it first in the web app to see if it fits our needs

PR #290, opened 2024-11-21
Add error monitoring section to privacy policy, dynamic confirmation popup in the ap

mclrc:


johan-t (commented):
Is it still fine if we report the data of the people who have not accepted the privacy policy? 

Because at least for the web app right now only people who have submitted a report have accepted the privacy policy.

johan-t:
Is there anything else we need to use sentry in a GdpR compliant way?

mclrc:
I configured sentry to not store IPs, we should be good on the GDPR front. I will add a privacy policy version file to public so that we can check in the app if the user confirmed the current version

johan-t:
@mclrc I mean as long as we dont store any data with which a person can be identified we should be good to go.

johan-t (approved):

johan-t:
@mclrc What does the dynamic privacy policy date solve or why did you choose to implement it?

mclrc:
@johan-t I implemented a dynamic blocking popup in the app that checks the confirmed and most recent privacy policy versions and forces users to accept the new one before using the app. this means we don't need to ship app updates (and have them reviewed + pushed out to all users) before we change something that requires a change to the privacy policy.

PR #317, opened 2024-11-28
chore: use ESLint config similar to app for frontend

mclrc:
## Describe the issue:
Previous eslint config was not very strict and inconsistent between app and website
used mostly the same eslint config for the website, except for react native specific rules
I disabled a11y (accessibility), no-console and a few other lints in a bunch of places for now because i felt fixing those issues would have exploded the scope of this PR, but it might be worth considering to address those issues at some point.

ALSO this isn't tested very well yet. Might be good if one of you could do that as I don't have the backend set up and also wouldn't know how to test the website too well


johan-t:
If there are too many issues that come up with the new config, you could configure the github action to only check modified files, so that we will fix the files one by one as we are working on them.

packages/frontend/.eslintrc.json
johan-t, line 1: just out of curiosity, why are we not using a global config?

packages/frontend/src/components/Buttons/CloseButton/CloseButton.tsx
```diff
@@ -4,6 +4,5 @@ interface CloseButtonProps {
     closeModal: () => void
 }
 
-export const CloseButton: React.FC<CloseButtonProps> = ({ closeModal }) => {
-    return <button className="small-button close-button center-child" onClick={closeModal}></button>
-}
+// eslint-disable-next-line jsx-a11y/control-has-associated-label, react/button-has-type
```
johan-t, line 7: I see that the react/button-has-type is being disabled a lot, why not disable it completely instead of for each line?

johan-t (commented):
Great to see this taking shape.

I am not sure if it is a good move to disable all of the rules that cause issues for the next line. As I find it hard to imagine that we will remove the //eslint-disable in the future and instead just stick with it because it is easier in the short term, meaning that the improved eslint config has little value. 
But I also agree that fixing each issue would be too much for one PR. 

Instead I would set up the CI so that it will only run linting for changes files. This way we can improve the code file by file without having one huge PR. This will also make it safer in case anything breaks as we can easily pin point the issue.

What do you think?

johan-t:
Also when changing something in the packages that more people are working on, please use commit message prefixes for the PR title, this will make it easier to understand what a developer changed in the future. For example this PR title could be: "chore: use ESLint config similar to app for frontend"

[This](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/) is a good article about the topic.

packages/frontend/.eslintrc.json
mclrc: lsp and module resolution issues, it was a headache trying to make everything resolve at all, i just decided to copy it. there are also some differences, the app has react native specific rules and i imagine there are some web specific ones that would make sense for the website

packages/frontend/src/components/Buttons/CloseButton/CloseButton.tsx
mclrc: because we should probably fix this for accessibility reasons

packages/frontend/.eslintrc.json
johan-t: Okay makes sense imo it is not that important that they match 100% I think it is fine if there are some smaller changes between them just that the general idea of the config is the same.

mclrc:
i would agree, but i don't think it's too much of a hassle to fix immediately. most of them are accessibility things or console uses, the former we can either choose to ignore or preferrably use semantically correct html and button types in the few instances where it is complaining, for the latter we should just replace almost every console.error with a sentry call and simply log those on dev. i'd even say that some of the others can remain as is, disabling a lint isn't necessarily a horrible thing, you just gotta be careful with it. we might not touch certain files in a long time, better to get it over with i think.

packages/frontend/src/components/Modals/MarkerModal/MarkerModal.tsx
```diff
@@ -117,9 +128,10 @@ const MarkerModal: React.FC<MarkerModalProps> = ({ className, children, selected
     const disclaimerWithLink = useMemo(
         () => (
             <>
-                <a href="#" onClick={handleShare} className="invite-link">
+                {/* eslint-disable-next-line react/button-has-type */}
+                <button onClick={handleShare} className="invite-link">
```
johan-t, line 132: Having a button instead of an `a` will break the ui due to default styles being applied to the button tag:

Bildschirmfoto 2024-11-28 um 16 15 33

packages/frontend/src/components/Map/Markers/Classes/OpacityMarker/OpacityMarker.tsx
```diff
@@ -1,11 +1,10 @@
-import React, { useEffect, useState, useMemo, useRef } from 'react'
+import './OpacityMarker.css'
```
johan-t, line 1: For some reason is the .pulse animation not applied. Not for the markers or for the new report button

packages/frontend/src/components/Modals/UtilModal/UtilModal.tsx
```diff
@@ -30,18 +31,22 @@ const UtilModal: React.FC<UtilModalProps> = ({ className, children, colorTheme,
...
                     <ul>
-                        <li onClick={toggleColorTheme}>
-                            {colorTheme === 'light' ? (
-                                <img src={light_icon} alt="Light Icon" />
-                            ) : (
-                                <img src={dark_icon} alt="Dark Icon" />
-                            )}
+                        <li>
+                            {/* eslint-disable-next-line react/button-has-type */}
+                            <button onClick={toggleColorTheme}>
```
johan-t, line 43: this will break the styling

johan-t (commented):
I found some styling issues. Note that I have not tested the risk model as it is not deployed for the test backend and there is no documentation for running it locally.

This kind of makes me less confident in deploying it all at once. I am sure that there are some issues I have overlooked or some that are extrem edge cases. What do you think @mclrc?

johan-t:
@mclrc I have merged #344 so you can close this PR. But thanks for the setup I just copied it.

Issue #300, opened 2024-11-24
Fix sorting order of stations in report modal


johan-t:
The stations are being sorted alphabetically based on their id, they should be sorted based on their name.

vinhyan:
Hi @johan-t! I can look into this issue.

johan-t:
> Hi @johan-t! I can look into this issue.

Awesome, let me know if you have any issues setting up the dev environment or with anything else

vinhyan:
Hi @johan-t! I was trying to get the backend to work and ran into the error below. Apparently `RISK_API_URL` env is needed here but I could not find this piece of information anywhere from the docs. Can you advise what this url should be?

<img width="1362" alt="Screenshot 2024-11-27 at 3 33 28 PM" src="https://github.com/user-attachments/assets/d5d63080-3af1-4310-a836-e8bd2be40609">

![image](https://github.com/user-attachments/assets/e558a260-29f3-4a1e-9c1f-e59ff30b7acd)

johan-t:
Hey, this is for our "microservice" that will calculate the risk for each line. It is a bit harder to get running and even harder to modify at the moment. For now you can ignore this error as the risk microservice is not needed for most issues.

vinhyan:
Hi @johan-t! 

Thank you for the clarifications. After studying the codebase, I have a couple of questions: 

As I set up my database locally (following the backend `README.md`), I started with an empty `ticket_info` table created by the script. When submitting a `report`, the `POST` route (`basics/inspectors`) checks the request `data` for hate speech using `checkHateSpeech`, which requires `CONTENT_MODERATION_SERVICE_URL`. Am I correct in assuming this URL refers to a "microservice" that is not publicly accessible, similar to `RISK_API_URL`? 

Currently, without `CONTENT_MODERATION_SERVICE_URL`, `PostInspector` is returning an `Internal Server Error`, which prevents generating any report data for use in `ReportsModal`.

![image](https://github.com/user-attachments/assets/7c0a9a7f-0c86-4722-975a-71025be26113)
![image](https://github.com/user-attachments/assets/a06f4f3f-df34-4e50-b088-c1c16f959adc)
![image](https://github.com/user-attachments/assets/e1134f38-4016-4c50-994b-c02bd09fb4bc)

I’m still familiarizing myself with the codebase, so there may be other areas I’ve missed. Any advice or guidance you can provide would be greatly appreciated.

Thank you!

johan-t:
[...]

PR #320, opened 2024-11-29
Sort stations by name in report form

vinhyan:
## Describe the issue:
Stations are not sorted by name when submitting a report. (issue #300)
In `ReportForm`, the list of stations is populated and stored in `possibleStations`. All stations are now sorted by name from the start (`sortedAllStations`).

When a `currentLine` is selected, stations of the line will be sorted since they are mapped from `allLines`.

When a `currentEntity` is selected or a station is searched, the list of stations is taken from `sortedAllStations`, which is already sorted.

N/A

### DIFF
--- a/packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
+++ b/packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
@@ -4,7 +4,7 @@ import { useTranslation } from 'react-i18next'
 import SelectField from '../SelectField/SelectField'
 import AutocompleteInputForm from '../AutocompleteInputForm/AutocompleteInputForm'
 
-import { LinesList, StationList, reportInspector } from '../../../utils/dbUtils'
+import { LinesList, StationList, reportInspector, StationProperty } from '../../../utils/dbUtils'
 import { sendAnalyticsEvent } from '../../../utils/analytics'
 import { highlightElement, createWarningSpan, getLineColor } from '../../../utils/uiUtils'
 import { calculateDistance } from '../../../utils/mapUtils'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
```diff
@@ -87,15 +87,31 @@ const ReportForm: React.FC<ReportFormProps> = ({ closeModal, notifyParentAboutSu
 
     // Calculate possible stations based on entity, line, station, and search input
     const possibleStations = useMemo(() => {
-        let stations = allStations
+        // Sort station records from a station record [stationKey, station] array by station name
```
johan-t: I would remove this comment, imo if you feel like you need a comment that explains the function then the function should be renamed to be self explanitory. But in this case I think the function name was chosen very well and it clearly explains what the function does, so no need for the comment.

packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
```diff
@@ -87,15 +87,31 @@ const ReportForm: React.FC<ReportFormProps> = ({ closeModal, notifyParentAboutSu
...
+        let stations = sortedAllStations
+
         if (currentStation) {
             stations = { [currentStation]: allStations[currentStation] }
         } else if (currentLine) {
             stations = Object.fromEntries(
-                allLines[currentLine].map((stationKey) => [stationKey, allStations[stationKey]])
+                allLines[currentLine]
+                    .map((stationKey) => [stationKey, allStations[stationKey]])
+                    .sort(sortStationRecordsByStationName)
```
johan-t: Once a line is selected, the stations naturally follow the train's traversal order along the line. I'm not entirely sure if people are faster if this order is disrupted, but I’d like to test this hypothesis. In my opinion, the best approach is to remove this line, merge and deploy your changes, and observe whether people can report inspectors more quickly. Once we’ve established a new baseline, we can reintroduce the line to test the hypothesis.

johan-t (changes requested):
Fantastic changes, I am so sorry about the unclear description of the issue and the misunderstanding. In the future if you want to solve a different issue let me know beforehand and I will write an issue description that is more clear.

Nevertheless thank you for taking the time to solve this issue. We had multiple people reaching out to us that they are annoyed by the order being wrong. So these changes will make the people very happy.

packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
vinhyan: Agreed!

packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
vinhyan: Oh that's such a great insight! I would have never known this. Personally, I also prefer keeping the station order as it is in real life. It makes more sense as I assume most riders using this app are regulars, and are familiar with that order. But please keep me posted if you ever decide to test it out.

vinhyan:
@johan-t Thanks for your feedback! Changes have been made in the new commit.

johan-t (approved):
lgtm 👍

packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
johan-t: will do

johan-t:
@vinhyan changes have been deployed. Thanks a lot for your help 🙏

PR #357, opened 2024-12-17
fix: avoid race conditions with funnel event

johan-t:
## Describe the issue:
We were trying to track how many users clicked on the link from the telegram group. This worked very seldomly.

I avoided the race condition by waiting for pirsch sdk to be loaded and only redirecting once this is done and the promise is returned.

Not 100% sure that this will work, lets see


johan-t:
IMG_1923

PR #365, opened 2024-12-22
feat: introduce server-side rate limiting

johan-t:
## Describe the issue:
There were some instances of people submitting a suspicious amount of reports. They were able to do this because there is no limit to how many reports are allowed from a single ip address. 

When there is a request to `POST /Inspectors` we are now checking if the ip address has reported in the last 30 minutes. If this is the case we are returning an error `429` indicating too many requests.
The ip addresses are being hashed using sha256 and salted. This will ensure an extra level of security and data privacy for the users.

I decided to keep the client side rate limiting in` ReportForm.tsx` as this will ensure that users cannot cheat the system by simply changing to mobile data or switching the wifi.

I also updated the analytics section of the pivacy policy to use the [example](https://docs.pirsch.io/privacy) provided by the pirsch docs. I believe this is safer and easier to handle then having to mention every event we are tracking.


packages/backend/limiting/limiting.go
```diff
@@ -0,0 +1,86 @@
...
+	hashedIP := r.hashIP(ip)
+	lastSubmission, exists := r.submissions[hashedIP]
+
+	if !exists {
+		return true
+	}
+
+	if time.Since(lastSubmission) > 30*time.Minute {
+		return true
+	}
```
mclrc, line 57: i'd probably define the timeframe in a constant somewhere

mclrc (approved):

PR #369, opened 2024-12-25
 feat: introduce server-side rate limiting #365 

johan-t:
## Describe the issue:
There were some instances of people submitting a suspicious amount of reports. They were able to do this because there is no limit to how many reports are allowed from a single ip address. 

When there is a request to `POST /Inspectors` we are now checking if the ip address has reported in the last 30 minutes. If this is the case we are returning an error `429` indicating too many requests.
The ip addresses are being hashed using sha256 and salted. This will ensure an extra level of security and data privacy for the users.
Requests from the telegram bot are now being made with a X-Password header, if a request has this header it is not effected by rate limiting. Otherwise we would rate limit the messages from the telegram group.

I decided to keep the client side rate limiting in` ReportForm.tsx` as this will ensure that users cannot cheat the system by simply changing to mobile data or switching the wifi.

I also updated the analytics section of the pivacy policy to use the [example](https://docs.pirsch.io/privacy) provided by the pirsch docs. I believe this is safer and easier to handle then having to mention every event we are tracking.


johan-t:
@brandesdavid I changed the .envs of the telegram bot. Do I need to change anything if I want to keep the dev docker enviroment in sync?

packages/backend/server/server.go
```diff
@@ -78,6 +79,14 @@ func SetupServer() *echo.Echo {
 		logger.Log.Error().Str("Error", err.Error()).Send()
 	}
 
+	_, err = c.AddFunc("* * * * *", func() {
+		limiting.GlobalRateLimiter.CleanupOldSubmissions()
```
brandesdavid, line 84: maybe add a note to the cronjob what it means. every month, week, day, hour and minutes it runs.

packages/frontend/public/locales/de.json
```diff
@@ -126,9 +126,10 @@
...
+                "purpose": "Die gemeldeten Daten werden anderen Nutzern der App temporär zur Verfügung gestellt, um ihnen die Möglichkeit zu geben, sich über die aktuelle Situation in den öffentlichen Verkehrsmitteln zu informieren. Des Weiteren werden die Informationen in die Freifahren_BE Telegram-Gruppe weitergeleitet. Diese Datenübermittlung dient dem Austausch und der Information über Vorfälle innerhalb der Community. Die IP-Adresse wird zur Erkennung von Spam- und Missbrauchsmeldungen verwendet um sicherzustellen, dass die Meldungen von echten Nutzern stammen und nicht von Bots oder anderen automatischen Quellen. Die IP-Adressen werden maximal {{minutes}} Minuten in stark verschlüsselter Form gespeichert und dann gelöscht. Sie sind nicht mit anderen Daten verknüpft und können nicht wieder auf den Nutzer zurückverfolgt werden."
```
brandesdavid, line 132: 👍

packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
```diff
@@ -239,14 +252,17 @@ const ReportForm: React.FC<ReportFormProps> = ({ closeModal, onNotifyParentAbout
...
-    
+
     const validateReportForm = async () => {
         let hasError = false
 
         // Check for last report time to prevent spamming
         const lastReportTime = localStorage.getItem('lastReportTime')
 
-        if ((lastReportTime !== null) && Date.now() - new Date(lastReportTime).getTime() < REPORT_COOLDOWN_MINUTES * 60 * 1000) {
+        if (
```
brandesdavid, line 262: what formatter are you using? we should all use the same one, for git blame and reduced changes in diff

brandesdavid (approved):
please answer some comments, merge if you want to

brandesdavid:
> @brandesdavid I changed the .envs of the telegram bot. Do I need to change anything if I want to keep the dev docker enviroment in sync?

no

packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
johan-t: I am using prettier with the config that is defined in root, for python I am using black with the default config. I was wondering if you have prettier installed or where the formatting changes are coming from.

PR #374, opened 2025-01-09
chore: add sentry

johan-t:
## Describe the issue:
We were having a tracking bugs, it mostly relied on internal testing. 

By adding sentry to the web app we can now monitor performance issues and track any uncaught errors. This will allow us to create a more stable app in the future and squash bugs easily.  
I added some custom error handling to the useAPI hook as API issues are the most critical to handle.
We can add some more custom error handling in the future how we see fit, I did not want to overdo it.

Not 100% sure that this will work in production.


johan-t:
Deployment is tricky due to react version missmatch. Could not fix until now

PR #380, opened 2025-01-20
feat: api versions

johan-t:
This is critical as more and more services rely on this backend. And the mobile apps new a review time until the changes will be rolled out and we need to account for users with older versions of the mobile app.

This means if the api contract would change the app would be broken until the users update it to the newest version. To prevent this we have api versions.


johan-t:
Once actual changes in the version are introduced we can add a better file structure for each endpoint. But I would roll this out one by one, so that we can adapt more easily to bad design decisions.

PR #396, opened 2025-02-07
feat: navigation microservice

johan-t:
## Describe the issue:
Every users has to go somewhere. We are in the best situation to provide them with that info and find the best way with the lowest risk.

To add this feature we need an API that we can request with a start and endstation and it should return the safest option plus some alternative routes.

I used the [transitious api](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/motis-project/motis/refs/heads/master/openapi.yaml#tag/routing) as a routing engine that our API can wrap around it.
I first created a mapping of the FreiFahren station ids to the engines station ids. We are then using this map to construct an api request to the engine based on the start and endstation property. 

Once we have received a response we find the the `Itinerary` with the lowest risk, by matching the `/segement-colors/` segments to the legs and add their risk value up.
Using this info we construct the return data with the safest route and 4 alternative routes that will be returned to the frontend.

- merged #387 into this branch


brandesdavid:
- Maybe add to the readme the needed env's
- and also add an example curl request and it's output

packages/navigation_service/src/server.ts
```diff
@@ -0,0 +1,305 @@
...
+        url.searchParams.set('arriveBy', 'false')
+        url.searchParams.set('timetableView', 'true')
+        url.searchParams.set('pedestrianProfile', 'FOOT')
+        url.searchParams.set('preTransitModes', 'WALK')
+        url.searchParams.set('postTransitModes', 'WALK')
+        url.searchParams.set('directModes', 'WALK')
+        url.searchParams.set('requireBikeTransport', 'false')
+
+        // Measure engine response time
+        const engineStartTime = Bun.nanoseconds()
```
brandesdavid: Is it possible to move some code for the measuring into another function? Like the construction of the performance metric struct and logging it could be moved. This would tidy the routing function more

brandesdavid (changes requested):
there some things you can add. but everything looks pleasing :)

packages/navigation_service/src/server.ts
johan-t: Where would you move it? The performance measurement cannot be moved to a different place otherwise it would measure something different, or am I not seeing something here?

johan-t: I just removed it

johan-t:
It would be nice if we could deploy this so that it is reachable under api.freifahren.org so that we reduce complexity for frontend developers, but it might not be worth it. Not sure how complex this would be

johan-t:
Rewrote this to be part of the main backend to avoid having a to make a request to a separate backend url, this would be add complexity for future maintainers and to avoid having to write a reverse proxy. This solution makes sure we keep things as simple as possible

johan-t:
I have decided to close this Pr and instead keep frontend and backend implementations in a single pr: #400 so that it is easier to changes to backend and frontend at the same time

PR #411, opened 2025-04-27
feat: support all trams + automatic data setup

johan-t:
## Describe the issue:
We had supported all metro trams, but not the non-metro trams. Even though reports on these lines are very rare, users kept requesting that we add them.

Adding new lines requires updating multiple files: `StationsList.json`, `LinesList.json`, `segments.json`, `synonyms.json`, and finally `stationsMap.prod.json`. Since there were about 20 new lines, doing this manually would have been far too tedious.

To solve this, I wrote a script that consumes some configurations for a city defined in `config.py` and automates the entire process. This means it is now possible to set up FreiFahren in a new city within 5 minutes.  
Note that some small Berlin-specific logic remains, and a few post-processing steps are still required.

- All existing station IDs were updated, meaning a database migration is required.
- I did not create synonyms for the new tram stations, as I thought it was not worth polluting the NLP models.


johan-t:
The test is failing as it takes data from the test database, which still has the old station ids, meaning it will only work after this PR is merged and the migrations are run.

packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
```diff
@@ -78,19 +78,28 @@ const ReportForm: FC<ReportFormProps> = ({ onReportFormSubmit, className }) => {
...
-                (accumulatedLines, [line, stations]) => ({
-                    ...accumulatedLines,
-                    [line]: stations,
-                }),
-                {} as LinesList
-            )
-    }, [allLines, currentEntity])
+        let filteredLines: [string, string[]][] = []
+
+        // 1. Filter by entity (U, S, M, etc.)
```
brandesdavid: Maybe also add a note here, that this is Germany based

scripts/create_lines_list.py
```diff
@@ -0,0 +1,132 @@
...
+from typing import Dict, List, Set, Tuple
+
+from config import LINES
+from geo import haversine
+
+STATIONS_JSON = "StationsList.json"
+OUTPUT_JSON = "LinesList.json"
+
+
+def build_mst(
```
brandesdavid, line 12: please provide a correctness proof for that
https://course.ccs.neu.edu/cs5002f18-seattle/lects/cs5002_lect11_fall18_notes.pdf

brandesdavid (approved):

packages/frontend/src/components/Form/ReportForm/ReportForm.tsx
johan-t: It's not really berlin based as even in other cities we would use the same entity code. I guess if people would configure them differently it would differ, but I'll add a note.

scripts/create_lines_list.py
johan-t: absolutely

PR #414, opened 2025-04-27
Station Name Routing

brandesdavid:
## Describe the issue:

We want to incentivize the users to click on the link, so they can find the report of the station directly on the map.

Added station page routing through /?stationId=(STATION-ID)
- Backend now sends the station id to the report inspector api of the telegram bot
- Telegram bot now sets the link to the app with /?stationId=(STATION-ID)
- The station select will be automatically opened, with the newest descriptions here.


packages/nlp_service/services/api_adapter.py
```diff
@@ -16,20 +16,20 @@ def report_inspector() -> tuple:
...
+    telegram_message = ""
 
-    telegram_message = "Über app.freifahren.org gab es folgende Meldung:"
-    telegram_message += "\n"
     telegram_message += f"\n<b>Station</b>: {station}"
     if line:
         telegram_message += f"\n<b>Line</b>: {line}"
     if direction:
         telegram_message += f"\n<b>Richtung</b>: {direction}"
     if message:
```
johan-t, line 31: We need to cover the case of in which message is None, otherwise it would be a step backwards

packages/nlp_service/services/api_adapter.py
```diff
@@ -16,20 +16,20 @@ def report_inspector() -> tuple:
...
-        telegram_message += f"\n<b>Beschreibung</b>: {message}"
+        telegram_message += f"\n<b>Beschreibung</b>: Auf app.freifahren.org/?stationId={stationId} findest du mehr Informationen"
```
johan-t: One thing I just realized is that on mobile the link is not that highlighted as it is in the middle of the sentence. I would format it like this instead:

"Einsehbar auf: app.freifahren.org?stationId="

And for no message something like this:

"Mehr informationen auf: app.freifahren.org?stationId="

PR #418, opened 2025-05-07
fix: tram detection issues

johan-t:
## Describe the Issue
Several issues arose due to the introduction of the new trams. Specifically, the logic was checking if a line was `in` a word rather than using `== word.lower()`. This led to cases where, for example, the M29 was incorrectly detected as the tram M2.  
Additionally, the U12 line was not being accounted for.

## How the Issue Was Resolved

- Ensured that a word must exactly match the line for detection.
- Numeric-only tram lines now require the previous word to be "tram" to avoid confusion with unrelated numbers.
- U12 is now detected as U2. I acknowledge that this is a suboptimal solution, as U12 can refer to either U1 or U2 depending on the station. However, since station detection currently depends on the line, this is a non-trivial issue. I'm open to suggestions for improving this logic.

### DIFF
--- a/packages/nlp_service/core/extractors/line_extractor.py
+++ b/packages/nlp_service/core/extractors/line_extractor.py
@@ -40,10 +40,6 @@ def format_text_for_line_search(text):
     return " ".join(formatted_words)
 
 
-"""
-Extraction functions
-"""
-
 def find_line(text, lines):
     logger.debug("finding the line")
 
@@ -57,7 +53,7 @@ def find_line(text, lines):
 
     for word in set(words):
         for line in sorted_lines:
-            if line.lower() in word.lower():
+            if line.lower() == word.lower():
                 matches_per_word.setdefault(word, []).append(line)
 
     return process_matches(matches_per_word)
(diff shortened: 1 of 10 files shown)


johan-t:
I will fix the tests later

packages/nlp_service/core/extractors/line_extractor.py
```diff
@@ -55,9 +52,16 @@ def find_line(text, lines):
...
         for line in sorted_lines:
-            if line.lower() in word.lower():
+            # For numeric-only lines, require "Tram" prefix
+            # as these lines are unlikely make sure the user is not referring to a number
+            if line.isdigit():
+                if i > 0 and words[i-1].lower() == "tram" and line == word:
+                    matches_per_word.setdefault(word, []).append(line)
+            elif line.lower() == word.lower():
                 matches_per_word.setdefault(word, []).append(line)
+            elif word.lower() == "u12": # temporary fix for u12, which is not in the lines dict
```
brandesdavid, line 64: Add a TODO, so we can remove this case, when the U12 is not operating anymore.

brandesdavid (approved):
lgtm

packages/nlp_service/core/extractors/line_extractor.py
johan-t: my idea was to keep this, because the U12 is reoccurring and it won't cause any harm if the U12 is gone, but that way we don't need to remember to add it again.

PR #419, opened 2025-05-10
Telegram mini web app to report inspectors fre 473

brandesdavid:
## Describe the issue:

The telegram group uses natural language to report, which is harder to get information from.  To provide more structured concise data and comfortable usage,
we have added a Telegram Mini App to the Telegram Bot, that reflects the FreiFahren WebApp

Added a new package `mini_app` as a react structure to deploy a mini app
- near replica of the current report form found in the webapp

Added an author id to inform the backend the report was made through the mini app.
Added an automatic /start command the telegram bot, to offer the service when a new chat has been started

This mini app has to be connected with the telegram bot and environment variables have to configured for production.


packages/mini_app/src/components/ErrorBoundary.tsx
johan-t, line 1: Do we really need an ErrorBoundary here? If we do not have a strong reason for adding it my personal move would be to not add it for now. The reason for this is that we should try to keep this as lean as possible to make it easier to maintain and not solve issues be do not have. But would love to hear your opinion :)

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
```diff
@@ -0,0 +1,496 @@
...
+                    )
+            )}
+        </div>
+    );
+};
+
+// Line component for display
+const Line: FC<{ line: string }> = ({ line }) => {
+    return (
+        <div className="line" style={{ backgroundColor: getLineColor(line) }}>
```
johan-t: Please make sure that the line is centered

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
johan-t, line 1: Can you remove the llm comments ;) and replace with comments when **You** think extra clarification is needed?

packages/mini_app/src/components/ReportSummaryModal/ReportSummaryModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+                </div>
+                <span className="user-count">
+                    <img
+                        className="users-icon"
+                        src="/users-svgrepo-com.svg"
+                        alt="users"
+                    />
+                    <h1>{animatedCount}</h1>
+                </span>
+                <p>Andere Nutzer haben heute bereits Kontrollen gemeldet. Gemeinsam sorgen wir für mehr Transparenz im öffentlichen Nahverkehr.</p>
```
johan-t: Can you use the same string as in the web app?

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
johan-t, line 1: Why is the component called `MAReportForm.tsx` what does "MA" stand for?

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
```diff
@@ -0,0 +1,496 @@
...
+                        <section>
+                            <div>
+                                {currentStation && (
+                                    <label>
+                                        <input
+                                            type="checkbox"
+                                            checked={isPrivacyChecked}
+                                            onChange={() => setIsPrivacyChecked(!isPrivacyChecked)}
+                                        />
+                                        Ich stimme der <a href="https://app.freifahren.org/datenschutz">Datenschutzerklärung</a> zu.
```
johan-t: can you add an indicator that this is required (the red *) like in the web app?

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
```diff
@@ -0,0 +1,496 @@
...
+            // Show error to user (could be enhanced)
+            if (error == "Error: HTTP error! status: 429") {
+                alert('Bitte nicht so schnell! Warte 30 Minuten und versuche es erneut.');
+            } else {
+                alert('Fehler beim Senden der Meldung. Bitte versuche es erneut.');
+            }
+        }
+    };
+
+    const handleCloseSummary = () => {
```
johan-t: Is it possible to close the web app from this function? As the user would probably not want to report a second inspector

packages/mini_app/src/utils/types.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+    } | null
+    line: string | null
+    isHistoric: boolean
+    message: string | null
+}
+
+// Added Position, Leg, Itinerary as they might be used by other copied parts indirectly
+// or could be useful for future extensions of the form.
+export type Position = {
```
johan-t: I would remove them if they are not used, we can add them as soon as we would use them

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
```diff
@@ -0,0 +1,496 @@
...
+                    coordinates: {
+                        latitude: 0, 
+                        longitude: 0 
+                    }
+                } : null,
+                message: descriptionRef.current?.value || null,
+                isHistoric: false
+            };
+            
+            await submitReport.mutateAsync(report);
```
johan-t: why did you use `mutateAsync` here?

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
```diff
@@ -0,0 +1,496 @@
...
+                                    ))}
+                                </SelectField>
+                            </section>
+                        )}
+                    </div>
+                    
+                    {/* Station selector */}
+                    <section className="station-selector">
+                        <h2>Station <span className="red-highlight">*</span></h2>
+                        <input
```
johan-t: Can you add a scroll effect in which if the user clicks this input the scroll pos. will be set so that the search is way at the top? This makes it a lot easier to see the possible stations on very small mobile devices.

johan-t: This is something that will also be added in the next iteration of the web app reportform

packages/backend/api/inspectors/postInspector.go
```diff
@@ -123,6 +124,19 @@ func PostInspector(c echo.Context) error {
 				logger.Log.Info().Msg("Skipping Telegram notification - rate limit not exceeded")
 			}
 		}
+
+		if pointers.AuthorPtr != nil && *pointers.AuthorPtr == 77105110105 {
```
johan-t: second check for `&& *pointers.AuthorPtr == 77105110105` is redundant

packages/nlp_service/services/api_adapter.py
```diff
@@ -44,4 +44,78 @@ def restart():
...
\ No newline at end of file
+        return {"status": "error", "message": "Invalid password"}, 401
+
+@flask_app.route("/mini-app", methods=["GET"])
+def serve_mini_app():
+    """Serve the Telegram Mini App HTML file"""
+    static_dir = os.path.join(os.path.dirname(os.path.dirname(__file__)), "static")
+    return send_from_directory(static_dir, "mini_app.html")
+
+@flask_app.route("/mini-app-report-inspector", methods=["POST"])
```
johan-t: can you use REST standard hierarchy so having the route like htis /mini-app/report

johan-t: also avoid using the term inspector as we are moving away from this for reasons with the mobile app not being able to call them that

johan-t: reports or report is always preferred

packages/nlp_service/services/api_adapter.py
```diff
@@ -44,4 +44,78 @@ def restart():
...
+        line = data.get("line", "")
+        station = data.get("station", "")
+        direction = data.get("direction", "")
+        message = data.get("message", "")
+        stationId = data.get("stationId", "")
+        
+        logger.info(f"Received Mini App data: {data}")
+        
+        # Generate a station ID (this would be replaced with your actual ID generation logic)
+        stationId = "mini-app-report"
```
johan-t: Why are replacing the stationId?

johan-t: please use the correct one

johan-t (changes requested):
functionality wise almost perfect, small modifications needed. Please go through the code again and remove the excessive LLM comments and replace them when they are needed. And make sure that the security issues are handled

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
brandesdavid: Mini App

packages/mini_app/src/components/ErrorBoundary.tsx
brandesdavid: deleted

packages/mini_app/src/components/ReportSummaryModal/ReportSummaryModal.tsx
brandesdavid: lame but ok

packages/mini_app/src/components/ReportForm/MAReportForm.tsx
johan-t: I would always avoid abbreviations, modern monitors are big and can support non abbrivated names

packages/nlp_service/services/api_adapter.py
```diff
@@ -44,4 +44,70 @@
...
+        message_text = request.json.get("message_text", "Click the button below to report inspectors:")
+        
+        # Send the Mini App button
+        send_webapp_button(chat_id, message_text, button_text, webapp_url)
+        
+        return {"status": "success"}, 200
+    
+    except Exception as e:
+        logger.error(f"Error sending Mini App: {str(e)}")
+        return {"status": "error", "message": str(e)}, 500
```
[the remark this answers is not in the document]
brandesdavid: No, the information will only be shown to the backend.

PR #420, opened 2025-05-12
align with latest docker compose file spec

greg6775:
As of Compose V2 the version tag is deprecated and the new recommended file name for the compose file is `compose.yaml`


johan-t:
Thanks for the suggestion!

PR #431, opened 2025-11-03
Partial initial setup

mclrc:


packages/hono-backend/src/app-type.ts
```diff
@@ -0,0 +1,7 @@
+export type Services = {
+  reportsService: ReportsService;
```
johan-t: ReportsService is undefined

packages/hono-backend/src/modules/reports/reports-service.ts
```diff
@@ -0,0 +1,18 @@
+import { DbConnection, reports } from "../../db";
+
+class ReportsService {
+  constructor(private db: DbConnection) {}
+
+  async getReports() {
+    const result = await this.db
+      .select({
```
johan-t, line 8: Does it make sense to specifically add a `Reports` type here that would be returned by the API? I know that it is already inferred but it might be nice to be extra specific here since it is used so much

johan-t: This is more of a nitpick so can be ignored

packages/hono-backend/src/db/schema/reports.ts
```diff
@@ -0,0 +1,20 @@
+import { pgTable, serial, integer, timestamp, pgEnum } from 'drizzle-orm/pg-core';
+import { createInsertSchema } from 'drizzle-zod';
+import { stations } from './stations';
+import { lines } from './lines';
+
+export const sourceEnum = pgEnum('source', ['mini_app', 'web_app', 'app', 'telegram']);
```
johan-t: nitpick: but I feel like "mobile_app" would be more specific instead of "app"

packages/hono-backend/src/db/schema/reports.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const reports = pgTable('reports', {
+  report_id: serial('report_id').primaryKey(),
+  station_id: integer('station_id').notNull().references(() => stations.station_id),
+  line_id: integer('line_id').references(() => lines.line_id),
+  direction_id: integer('line_id').references(() => lines.line_id),
+  timestamp: timestamp('timestamp').notNull().defaultNow(),
+  source_id: sourceEnum('source_id').notNull(),
```
johan-t, line 14: On second thought, do we need the source in the DB? Since technically we do not need to read it ever, we just need it for processing the request, which we can achieve by accepting it via HTTP but not storing it. I feel like it does not really hurt storing it, but it might also be better to keep the schema as lean as possible and avoid storing things that can be easily derived. 
What do you think?

johan-t (commented):
I really like the structure that this adds 👍. Excited to get it merged, the only thing blocking for me is the missing import of the type. The rest are things I just wanted to point out in case you think they are helpful for maintenance

packages/hono-backend/src/modules/reports/reports-service.ts
mclrc: we should definitely do this, but it should be inferred from the parsers in the front-end (or a shared package, ideally). doesn't really add anything to be specific here if the client doesn't use the same type

packages/hono-backend/src/db/schema/reports.ts
mclrc: I guess it doesn't really matter, but I do think it would be nice to have an overview over what share of reports comes from where. for the front end, we can do this with our analytics solution, but it would be weird to dispatch analytics events from the telegram bot, no?

johan-t: true

PR #434, opened 2025-11-07
Finish DB/local testing setup

mclrc:
I finished the DB setup including the missing fields, and added a basic compose setup for local testing that includes a Postgres DB. 
I also added a seed script that loads the stations and lines into the DB. 
We can delete/adjust a lot of this before deploying to prod, but I think for now it's quite useful.


packages/hono-backend/src/db/seed/StationsList.json
johan-t, line 1: I think it might be better if we don't manage the station ids by ourself and instead let the DB create them. This would mean less maintenance for us. Same might be good for line id, although fixed line ids might make sense so that you don't have to look up the name every time since the id is readable. What do you think?

johan-t (commented):
Very good changes. The only thing I am unsure about is if we want to manually maintain the ids. 

Also can you add some documentation in the `README.md`, about how this project can be set up? Just a very basic run down of what to set as `DATABASE_URL` and how to run the db via docker and that a new user has to seed the db.

packages/hono-backend/compose-hono.test.yaml
```diff
@@ -0,0 +1,38 @@
+services:
+  postgres:
+    image: postgres:16-alpine
```
johan-t: Why did you choose Postgres 16?

packages/hono-backend/src/db/seed/StationsList.json
mclrc: we should do this in production, but for tests (and local testing in the early stages) I think it's very useful to specify them yourself

packages/hono-backend/compose-hono.test.yaml
mclrc: claude did lol. will update

packages/hono-backend/src/db/seed/StationsList.json
mclrc: Also, the ids are very useful for inserting and the relations. otherwise, we wouln't be able to mass insert all the data correctly and have the relations resolve like this. we would have to query in between and save the db generated ids for each row to populate the next table

PR #435, opened 2025-11-07
FRE-538 + FRE-533 Post report

mclrc:
I'll write something here later


mclrc:
I think we might want to unify the station/lines services into a TransitNetworkData or similar service because it can be a little hard to figure out where logic should live and they will probably both stay relatively lean if kept separate. what do you think?

johan-t:
> I think we might want to unify the station/lines services into a TransitNetworkData or similar service because it can be a little hard to figure out where logic should live and they will probably both stay relatively lean if kept separate. what do you think?

That seems like a good idea. Do you think we should also keep the segments in the same service? I think they share a lot of the functionality for caching etc.

packages/hono-backend/package.json
```diff
@@ -12,9 +12,13 @@
     "db:seed": "bun run src/db/seed/index.ts"
   },
   "dependencies": {
+    "@types/lodash": "^4.17.20",
+    "@types/luxon": "^3.7.1",
```
johan-t: Do we really need luxon?

packages/hono-backend/src/modules/reports/reports-service.ts
```diff
@@ -1,18 +1,95 @@
...
+            lte(reports.timestamp, to.toJSDate()
+          )
+        )
+      );
 
     return result;
   }
+
+  async createReport(reportData: InsertReport) {
+    const postprocessed = await this.postprocessReport(reportData);
```
johan-t: If I remember we discussed briefly if we want to have a validator method similar to the handler method in `*-routes.ts`. Would it make sense to move this post processing in to that method?

packages/hono-backend/src/modules/reports/reports-service.ts
```diff
@@ -1,18 +1,95 @@
...
+
+    const result = await this.db
+      .insert(reports)
+      .values(postprocessed)
+
+    return result;
+  }
+
+  private async postprocessReport(reportData: InsertReport) {
```
johan-t: There are a bunch of cases missing in this. The post processing can become pretty complex because it has a lot of different cases that are also depending on each other. Also it is very important that this part is being properly tested. I think its best if you remove the current post processing so that we can address it as a separate issue and take enough time for it.

packages/hono-backend/src/modules/stations/stations-service.ts
```diff
@@ -0,0 +1,21 @@
+import { eq } from "drizzle-orm";
+
+import { DbConnection, lines, lineStations } from "../../db";
+
+export class StationsService {
+  constructor(private db: DbConnection) {}
+
+  async getLinesForStation(stationId: string) {
```
johan-t: Right now this is fine, but we have to think about caching for this. Utility functions like this are going to be called a lot and it is not worth it to make a new db transaction for each call of the function since the data is static and we can keep a copy of the table in memory at all times. So we have to think of a nice caching layer on top of the static tables, the best solution would be something like tanstack does where the developer does not even notice that the request is being cached.

packages/hono-backend/src/modules/reports/reports-routes.ts
```diff
@@ -1,12 +1,61 @@
...
 
-    return c.json(await reportsService.getReports());
+    const query = c.req.valid('query');
+
+    const range = {
+      from: query.from ?? DateTime.now().minus({ minutes: 59 }),
+      to: query.to ?? DateTime.now()
+    };
+
+    if (range.to.diff(range.from).as('hours') > 1) {
```
johan-t: This is incorrect as we should support queries for more than 1 hour. We do it for the statistics for example

johan-t: Instead we should return a bad request if the diff is smaller than 0, so if the "from" is after the "to"

johan-t: Also I feel like for this a validator method would be nice so that we separate the concerns from the handler

packages/hono-backend/src/modules/reports/reports-routes.ts
```diff
@@ -1,12 +1,61 @@
...
+  },
   handler: async (c) => {
     const reportsService = c.get('reportsService');
 
-    return c.json(await reportsService.getReports());
+    const query = c.req.valid('query');
+
+    const range = {
+      from: query.from ?? DateTime.now().minus({ minutes: 59 }),
+      to: query.to ?? DateTime.now()
```
johan-t: Testing this would be pretty nice

packages/hono-backend/src/modules/reports/reports-routes.ts
```diff
@@ -1,12 +1,61 @@
...
+    if (range.to.diff(range.from).as('hours') > 1) {
+      return c.json({ message: "Bad request - range larger than 1h" }, 400);
+    }
+
+    return c.json(await reportsService.getReports(range));
+  }
+})
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const getReportSource = (_: Context): InsertReport['source'] => 'web_app';
```
johan-t: Does it make sense to default to web_app here? Since most reports are from the bot, would it make sense to use that as the default?

johan-t (changes requested):
Nice PR. Some things I have noticed while testing:
- would be nice to include the correct DATABASE_URL env in the Readme or to have .env.example
- Adding the drizzle studio to just makes testing also a lot easier

In general I think it is better to create one PR for one issue (if possible), I know that it is annoying since things are interdependent, but otherwise we get into situations in which PRs get too big when doing the full scope. For example adding tests and complete validation logic will make this PR huge, this creates a slow down and makes things harder to review.
The alternative of doing the stuff like testing later is also sub optimal because we loose context about the implementation when doing it later, this makes it easy to forget some cases that should be tested. Or in other words it just generates technical debt that slows us down.

packages/hono-backend/package.json
mclrc: trust me working with dates sucks without something like luxon, so yes i'd say we definitely need it.

packages/hono-backend/src/modules/reports/reports-service.ts
mclrc: the cases that are missing are guessing the station when it's missing, which i think we should talk about again anyway. should be easy to add though.

packages/hono-backend/src/modules/reports/reports-service.ts
mclrc: postprocessing is pretty squarely business logic and should live in the service i think. also needs db access.

packages/hono-backend/src/modules/stations/stations-service.ts
mclrc: drizzle supports transparent caching, didn't set it up in here but should be pretty easy

packages/hono-backend/src/modules/reports/reports-routes.ts
mclrc: i think in the ticket it said to have a maximum time span here. might not have been 1h but 5. stats should get their own endpoint anyway, no?

mclrc: i can move it to the schema, it's not really schema validation but i guess it's fine. will see how readable it is.

packages/hono-backend/src/modules/reports/reports-routes.ts
mclrc: just hard coded it for now. didn't know what the different user agents/source detection criteria were

mclrc:
> Nice PR. Some things I have noticed while testing:
> 
>     * would be nice to include the correct DATABASE_URL env in the Readme or to have .env.example
> 
>     * Adding the drizzle studio to just makes testing also a lot easier
> 
> 
> In general I think it is better to create one PR for one issue (if possible), I know that it is annoying since things are interdependent, but otherwise we get into situations in which PRs get too big when doing the full scope. For example adding tests and complete validation logic will make this PR huge, this creates a slow down and makes things harder to review. The alternative of doing the stuff like testing later is also sub optimal because we loose context about the implementation when doing it later, this makes it easy to forget some cases that should be tested. Or in other words it just generates technical debt that slows us down.

i agree in general, i think in this case it's a pretty reasonable scope but i'll try to be more strict about it in the future lol. about the DB url: there is no need to set anything right now, it's handled by compose, that's why i didn't document it. once it's actually ready to deploy and use a real db, i agree we should doc it. i agree about the tests too, but we haven't really decided on a testing strategy/setup yet.

mclrc:
i can split the post-processing into it's own PR with tests. but lets maybe talk soon about how we want to do testing, which tools/kinds of tests we want etc.

packages/hono-backend/package.json
johan-t: Okay 👍

packages/hono-backend/src/modules/reports/reports-service.ts
johan-t: From the top of my head I think the cases missing are:
- guessing station
- inverting direction
- setting line based on direction
There might be some more.

Guessing station is very important in my opinion since the bot does not detect the station pretty often. If we don't guess the station we are loosing the report completely, but we can talk about it a bit more to make a final decision.

packages/hono-backend/src/modules/stations/stations-service.ts
johan-t: Very nice

packages/hono-backend/src/modules/reports/reports-routes.ts
johan-t: Personally I would not add a new endpoint for stats, since it is pretty much the same logic. Adding a new endpoint would:
- increase API complexity for developers
- add more code to maintain

Tbh, on second thought I don't think that we need to check the timeframe here, since we have a low timeout (5s) set for the DB anyways, meaning this is already handled by the DB (Which is the better solution since it fits all endpoints). I cannot think of a different reason why we would need a max timeout other than not blocking the db for too long. What do you think?

johan-t:
> i can split the post-processing into it's own PR with tests. but lets maybe talk soon about how we want to do testing, which tools/kinds of tests we want etc.

That sounds like a good idea.

johan-t:
> > Nice PR. Some things I have noticed while testing:
> > ```
> > * would be nice to include the correct DATABASE_URL env in the Readme or to have .env.example
> > 
> > * Adding the drizzle studio to just makes testing also a lot easier
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In general I think it is better to create one PR for one issue (if possible), I know that it is annoying since things are interdependent, but otherwise we get into situations in which PRs get too big when doing the full scope. For example adding tests and complete validation logic will make this PR huge, this creates a slow down and makes things harder to review. The alternative of doing the stuff like testing later is also sub optimal because we loose context about the implementation when doing it later, this makes it easy to forget some cases that should be tested. Or in other words it just generates technical debt that slows us down.
> 
> i agree in general, i think in this case it's a pretty reasonable scope but i'll try to be more strict about it in the future lol. about the DB url: there is no need to set anything right now, it's handled by compose, that's why i didn't document it. once it's actually ready to deploy and use a real db, i agree we should doc it. i agree about the tests too, but we haven't really decided on a testing strategy/setup yet.

But is there a solution how you can use the db studio with just? The only way I managed to get it running was to start the db via postgres and then run `bun run db:studio` and `bun run dev`, which works but is not that nice since we are not using just for it.

johan-t:
@codex review this PR.

packages/hono-backend/src/modules/reports/reports-routes.ts
mclrc: I definitely think we need a limit here, it can be large like 1W but it should be there. we don't want to hog the DB with a massive query we're not going to use

johan-t: In the end I don't think that it is very important, since the db timeout takes care of the issue of hogging the db for too long. If the time frame is loo large the db will reject the request and a 500 error will be returned automatically. But we can also specify a timeframe if you think it is necessary

johan-t (approved):

PR #437, opened 2025-11-11
feat: `GET transit/stations`

johan-t:
## Description:
I added the endpoint: `api.freifahren.org/transit/stations/` It returns the entire stations data in this format:
```
type StationId = string
type LineId = string

type Stations = Record<StationId, Station>
type Station = {
    name: string
    coordinates: { latitude: number; longitude: number }
    lines: LineId[]
}
```
I also moved the endpoint to be a sub route of `/transit/` so that the `getStations` and `getLines` can be handled in one service. Otherwise we the service structure would drift from the API structure. Let me know what you think of this.

## Implementation
The entire serialization step will be done at database level in a single transaction. This ensures we have maximum performance. It might be a little over the top since we will cache this any ways in the future (Meaning the db calls runs only once). Let me know if I should revert to the state from: [65141c2](https://github.com/FreiFahren/FreiFahren/pull/437/commits/65141c2db88f623f5b6abdc5d8cecf0753cdbfe4) since that implementation is slower but easier to read.

I would also like to get some feedback on how the types are created. @mclrc is there a way how we can a better job at inferring this from some db schema?

## Additional Notes
In order to not bloat this PR I will:
- add server side caching in a later PR
- add tests (not sure if we need tests for this endpoint, would like some feedback)


johan-t:
@codex review this PR. Take the description into account

packages/hono-backend/src/modules/transit/transit-network-data-service.ts
```diff
@@ -0,0 +1,48 @@
...
+                    lng: stations.lng,
+                    lineId: stationLines.lineId,
+                })
+                .from(stations)
+                .leftJoin(stationLines, eq(stationLines.stationId, stations.id))
+        })
+
+        const result: Stations = {}
+
+        for (const row of joinedRows) {
```
mclrc: for-of has a runtime penalty, not as bad in bun as other runtimes AFAIK but still. generally I think we should stick to functional/immutable patterns, reduce would probably work in this case. that should also take advantage of intrinsic JSC optimizations better. there is a lint for this as well.

packages/hono-backend/src/modules/transit/transit-network-data-service.ts
```diff
@@ -0,0 +1,48 @@
...
+type Station = {
+    name: string
+    coordinates: { latitude: number; longitude: number }
+    lines: LineId[]
+}
+
+export class LinesStationService {
+    constructor(private db: DbConnection) {}
+
+    // Perform entire query in a single transaction to get maximum performance by serializing at DB level
```
mclrc: are you sure about this? totally possible but AFAIK transaction doesn't do much for a single query, and the transaction start and commit have a runtime cost.

packages/hono-backend/src/modules/transit/lines-station-service.ts
```diff
@@ -0,0 +1,48 @@
...
+type LineId = string
+
+type Stations = Record<StationId, Station>
+type Station = {
+    name: string
+    coordinates: { latitude: number; longitude: number }
+    lines: LineId[]
+}
+
+export class LinesStationService {
```
mclrc: so we are unifying the lines/station services? definitely works for me, I might prefer a name like "TransitNetworkService" or something. that would fit other functionality that would be well placed here better, but lmk your thoughts on this.

packages/hono-backend/src/modules/transit/transit-network-data-service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+type StationId = string
+type LineId = string
+
+type Stations = Record<StationId, Station>
+type Station = {
+    name: string
+    coordinates: { latitude: number; longitude: number }
+    lines: LineId[]
+}
```
mclrc, line 15: would be a slightly different structure, but something like
```ts
export type Station = InferSelectModel<typeof stations> & {
  lines: InferSelectModel<typeof lines>["id"][];
};
```
could work.

packages/hono-backend/src/modules/transit/lines-station-service.ts
johan-t: What about "TransitNetworkDataService", I think "TransitNetworkService" could cause confusions once we add the transit service for routing

packages/hono-backend/src/modules/transit/transit-network-data-service.ts
johan-t: yeah no, i had some misunderstanding here about how drizzle works. I removed the explicit transaction since we are using a single query and it is not needed here.

johan-t: but postgres wraps every query in a transaction anyways, but the transaction function in drizzle is used to wrap multiple queries as far as i understand. So removing it is more straight forward even if it has the same performance

mclrc (approved):

PR #438, opened 2025-11-14
feat: test `from` and `to` query param logic

johan-t:
## Describe the issue:
We implemented the `from` and `to` query params without testing it.

I added tests that will test the following cases:
- `returns data in the specified timeframe when from and to query params are present`
- `returns data in the standard timeframe when query params are missing`
- `returns 400 when from is after to`
- `returns 400 when timeframe is longer than the maximum'`

I also refactored the report service to make use of some constants, so that these constants can also be used in the test. This will make sure that when we change the default timeframe we won't have to update the test with it.

All of this will now be run in CI, where we are testing against a proper test db.

- I know we decided that we wanted to mock the DB, but we decided to do this because we wanted a solution that is as simple as possible. While implementing I realized that proper db mocks are much more complex than creating a new test db. That is why I created the file `01_create_dbs.sql`, which will be run by postgres automatically. This creates the test db which is a copy of the freifahren db.  Let me know what you think?
- I had to alter the creation of the db to use the new format since we do not want to use migrations until release
- I removed the go test CI since it kept breaking due to the recent change and it is not needed anymore.


mclrc:
ah yeah I can see how mocking would be ugly for this case. I do still think we should consider mocking first for queries without a where clause like this, makes for more lightweight tests, and has the test case data right there in the test itself.

.github/workflows/run-hono-backend-CI.yml
```diff
@@ -27,3 +27,59 @@ jobs:
...
             - name: Run Prettier check
               run: cd packages/hono-backend && bun run format:check
+
+    test:
+        runs-on: ubuntu-latest
+        needs: lint
+        permissions:
+            contents: read
+        env:
+            DATABASE_URL: ${{ secrets.HONO_TEST_DB_URL }}
```
mclrc: can't we just run the tests in docker and save us the trouble of managing the GH secrets?

mclrc: If we're seeding it to the same state every time anyway (which we should do IMO) why not just use a containerized instance?

johan-t:
> ah yeah I can see how mocking would be ugly for this case. I do still think we should consider mocking first for queries without a where clause like this, makes for more lightweight tests, and has the test case data right there in the test itself.

So from what I read about this topic, it seems like the best practice is: mock the db when writing unit tests, use a real db when writing integration tests. I think we should stick to this pattern, but I would still use integration tests as a default and defer to unit tests if it makes sense for the situation.

.github/workflows/run-hono-backend-CI.yml
johan-t: good idea, I changed it 👍

johan-t:
@codex review this PR

mclrc (approved):

PR #440, opened 2025-11-16
feat: notify telegram about new report + error handling

johan-t:
## Describe the issue:
After each new report created either from the web app or the mobile app, we should send a request to the Telegram bot so that it can post a message in the Telegram chat.
Additionally, we previously had no error handling in place.

### Telegram Notification 
When an insert into the database is successful, we now send a `POST` request to the Telegram bot to notify it about the new report. To accomplish this, I enabled the source field to be sent via HTTP. This approach naively trusts the API user to provide the correct source (I added a test case to ensure it still defaults to telegram).
The alternative would be to parse the user agent on the backend to determine the source without user input. However, this approach is trickier to implement reliably, so I propose we stick to the naive version for now until it becomes insufficient.

The way we currently construct the API request for Telegram is a bit odd. I added a comment to remind us that this should be improved in the long run.

### Error handling
Since this feature requires interaction with multiple services, it is more prone to failure than simpler functionality. To mitigate this, we need proper error handling. In this PR, I introduced a general structure for handling HTTP errors, inspired by the way Go deals with errors (a method that may not look pretty but encourages developers to treat safety as a first-class concern).

The basic idea is that our service function returns an error of type `ServerErrorBody`:
```
export interface ServerErrorDetails {
    internal_code: InternalCode
    description?: string
}

export interface ServerErrorBody {
    message: string
    details: ServerErrorDetails
}
```
The route function checks whether an error occurred. If so, it logs the error and returns it. Otherwise, it responds with `status 200`:
```
        const error = await reportsService.createReport({
            ...reportData,
            source: reportData.source ?? 'telegram',
        })

        if (error !== null) {
            console.error(`Failed to create report: ${error.details.description}`)
            return c.json(error, 500)
        }

        return c.json(
            await reportsService.getReports({
                from: DateTime.now().minus({ hours: 1 }),
                to: DateTime.now(),
            })
        )
```

The purpose of this error type is to provide an `internal_code` that uniquely identifies the error more specifically than standard HTTP status codes like `500`. We can use this code on the client side to match and display a user-friendly error message.
We also provide a `message` and a `description`: the message serves as a heading, while the description gives additional context, mainly for developers to debug issues more easily.

This level of detailed error handling is necessary due to recent user feedback, as some users complained that it was not transparent why a report failed.


packages/hono-backend/src/http-error.ts
```diff
@@ -0,0 +1,11 @@
+export type InternalCode = 'TELEGRAM_NOTIFICATION_FAILED'
+
+export interface HttpErrorDetails {
```
brandesdavid: Naming from Http to ServerErrorDetails. I had the misconception it is like internal code 503 bad gateway

johan-t:
@codex review this PR

packages/hono-backend/src/modules/reports/reports-routes.ts
```diff
@@ -49,25 +48,29 @@ export const getReports = defineRoute<Env>()({
...
+        }),
     },
     handler: async (c) => {
         const reportsService = c.get('reportsService')
 
         const reportData = c.req.valid('json')
 
-        await reportsService.createReport({
-            source: getReportSource(c),
+        const error = await reportsService.createReport({
```
mclrc: I like errors as values as much as the next guy, but it's unfortunately not how JS works. I'd stick to more verbose try/catch, with this approach we trust that this function never throws for our error handling to work, which is a hard assumption to verify if we use this approach throughout the codebase.

mclrc (commented):
Regarding the error handling: I'd be careful about leaking errors to the client and the errors-as-values approach. I think we could probably create a `ClientError` class that we can use to describe errors we need to communicate to the client (without any more details than necessary or any data about the error object that was actually thrown) and then create a hono middleware that replaces any non-`ClientError` that occurs with a generic 500 client error and sends that to the client instead. for dev it might be useful to show specifically what happened but for prod I'd avoid it. what do you think?

packages/hono-backend/src/modules/reports/reports-routes.ts
johan-t: Good call, I refactored it now so that we can throw an error like this inside of a `try` `catch`:
```
throw new AppError({
                    message: 'Notifying the telegram bot went wrong',
                    internalCode: 'TELEGRAM_NOTIFICATION_FAILED',
                    description: `Failed to notify Telegram bot about inspector report: ${reason}`,
                })
```

If the developer does not manually throw an error and provide details it will default to something like this in dev:
```
{
    "message": "Internal Server Error",
    "details": {
        "internal_code": "UNKNOWN_ERROR",
        "description": "Cannot read properties of undefined (reading 'id')"
    }
}
```
and like this in prod:
```
{
    "message": "Internal Server Error",
    "details": {
        "internal_code": "UNKNOWN_ERROR"
    }
}
```
This is done via a global error handler that catches unhandled errors and formats them depending on env, this avoids that we leak any sensitive data to the client.

johan-t:
@codex review the new error handling

packages/hono-backend/src/modules/reports/reports-service.ts
```diff
@@ -24,7 +35,70 @@ export class ReportsService {
...
-        return this.db.insert(reports).values(reportData)
+    async createReport(reportData: InsertReport): Promise<void> {
+        await this.db.insert(reports).values(reportData)
+
+        if (reportData.source !== 'telegram' && process.env.NODE_ENV === 'production') {
+            try {
+                await this.notifyTelegram(reportData)
+            } catch (error) {
+                const reason = error instanceof Error ? error.message : 'Unknown error'
+                throw new AppError({
```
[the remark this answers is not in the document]
johan-t: We are now returning a header in this case

PR #443, opened 2025-12-04
Add analytics script in html

brandesdavid:
## Describe the issue:

Need a new analytics

Setted up analytics container. 

Exchanged pirsch with our script html. 

Updated the OptIn/Out Analytics option to use umami's mechanism

Updated privacy Policy page


packages/frontend/public/locales/de.json
```diff
@@ -181,8 +181,8 @@
...
-                "usage": "Diese Daten sind nicht personenbezogen; es werden keine Informationen gesammelt, die eine Identifizierung der Nutzer ermöglichen..."
+                "content": "Wir verwenden optional Umami Analytics für anonyme Nutzungsstatistiken und hosten es bei uns intern. Umami ist eine cookiefreie, Open-Source-Webanalyselösung und ist so konfiguriert, dass sie datenschutzfreundlich arbeitet. Die Erhebung erfolgt nur nach ausdrücklicher Zustimmung des Nutzers; ohne diese Einwilligung findet keine Analyse statt.",
```
johan-t, line 184: "Erhebung erfolgt nur nach ausdrücklicher Zustimmung des Nutzers; ohne diese Einwilligung findet keine Analyse statt" is this true?

johan-t: The way we handled this in pirsch was that we could collect data without the users approval since it was so anonymized. For the web app it would be cool if we can continue with this but for me personally not a hard requirement. @mclrc what do you think about this?

mclrc: I'd say we always want to collect. should be just as anonymized if we don't use the identify feature. but the web version might automatically do some fingerprinting, I don't know. for mobile I won't do this and we can keep going like we have been

PR #446, opened 2025-12-11
feat: post process report

johan-t:
## Describe the issue:
Sometimes either users or the bot submit reports that are:
1. incomplete 
2. impossible
We need to validate those, to make sure that we only have valid and as complete as possible reports in the database.

I defined a pipe that calls a couple of cross validation function on the report that was being sent by the user. 

A lot of this is coming from the extensive testing. If you have any ideas what we should also test in this context please let me know.
I also tried to write this as functional as possible so that we can reduce complexity.

### Important!!!
`guessStation` is not scaleable. It will fetch all of the reports for the line into memory and choose the most common in memory. This means as we get more reports the memory (and also CPU) usage of posting an inspector will increase. 
There are a couple of ways how we can solve this each with their own trade off and I would like some opinions on what we should do:
1. <del>Move the entire calculation to the DB. We can craft one drizzle query that will perform this all on the db side. The benefit of this is that this is as performant as it can possibly get since the brightest minds in computer science are working to optimize postgres and no matter how hard we try we will not be faster than Postgres. The downside of this is that the query would be quite complex, meaning difficult to write and even more difficult to read. </del>
2. <del>Create DB aggregates. We can create some aggregates that store the most common station for each line for each day of the week and each hour. We can then update this with a trigger every 24h at midnight or something like this. This would mean we have a O(1) lookup time but we also increase complexity of the db schema significantly. The benefit of this is that we can reuse the same logic for historic data, which means that the `POST /reports/` and the `GET /reports/` both benefit from this.</del>
3. Limit results: We can simply return the results of the last 30 or 90 days, with which we would have a very simple solution that is also very fast. The issue with this is that we loose significant accuracy especially if we would deploy in a city with a less active community.
4.<del> Do nothing. We currently have around 80k reports, even if all of them are loaded into memory it would not be that big of deal especially since we are not getting that many reports concurrently the server can probably handle it for now. So we could just note this down and solve the issue once it becomes an actual issue.  </del>

Also depending on the decision I can make `guessStation` be pure.

### DIFF
--- a/packages/hono-backend/src/common/utils.ts
+++ b/packages/hono-backend/src/common/utils.ts
@@ -1,0 +1,12 @@
+import type { Stations, StationId } from '../modules/transit/types'
+
+const lookupStation = (
+    stations: Stations,
+    stationId: StationId | null | undefined
+): Stations[StationId] | undefined => {
+    if (stationId === null || stationId === undefined) return undefined
+    if (!Object.prototype.hasOwnProperty.call(stations, stationId)) return undefined
+    return stations[stationId]
+}
+
+export { lookupStation }
--- a/packages/hono-backend/src/modules/reports/post-process-report.ts
+++ b/packages/hono-backend/src/modules/reports/post-process-report.ts
@@ -145,7 +145,9 @@ const correctDirectionIfImplied =
         const stationIndex = stationsOnLine.indexOf(reportData.stationId)
         const directionIndex = stationsOnLine.indexOf(reportData.directionId)
 
+        // If the station or direction is not on the line, return the report data unchanged
         if (stationIndex === -1 || directionIndex === -1) return reportData
+        // If the station and direction are the same, return the report data unchanged
         if (stationIndex === directionIndex) return reportData
 
         const impliedTerminalDirectionId = stationIndex < directionIndex ? lastStationOnLine : firstStationOnLine
--- a/packages/hono-backend/src/modules/reports/post-process-report.ts
+++ b/packages/hono-backend/src/modules/reports/post-process-report.ts
@@ -25,14 +25,6 @@ const pipeAsync = async <T>(value: T, ...fns: Array<(arg: T) => T | Promise<T>>)
     return result
 }
 
-const normalizeDayOfWeek = (dayOfWeek: number): number => {
-    // Accept both JS-style (0-6, Sunday=0) and Luxon-style (1-7, Monday=1)
-    if (!Number.isFinite(dayOfWeek)) return 0
-    if (dayOfWeek >= 1 && dayOfWeek <= 7) return dayOfWeek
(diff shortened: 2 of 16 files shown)


johan-t:
@codex review this PR

johan-t:
Todo: add timeframe filtering for guessStation and make it pure

packages/hono-backend/src/common/utils.ts
```diff
@@ -0,0 +1,12 @@
+import type { Stations, StationId } from '../modules/transit/types'
+
+const lookupStation = (
+    stations: Stations,
+    stationId: StationId | null | undefined
+): Stations[StationId] | undefined => {
+    if (stationId === null || stationId === undefined) return undefined
+    if (!Object.prototype.hasOwnProperty.call(stations, stationId)) return undefined
```
brandesdavid, line 8: Do you think we should return an error if the station is undefined?

brandesdavid: Or log it

packages/hono-backend/src/modules/reports/post-process-report.ts
```diff
@@ -0,0 +1,268 @@
...
+    let result = value
+    for (const fn of fns) {
+        result = await fn(result)
+    }
+    return result
+}
+
+const normalizeDayOfWeek = (dayOfWeek: number): number => {
+    // Accept both JS-style (0-6, Sunday=0) and Luxon-style (1-7, Monday=1)
+    if (!Number.isFinite(dayOfWeek)) return 0
```
brandesdavid: Is this function necessary? We are only using Luxon DateTime and could just assert if the weekday is 0.

packages/hono-backend/src/modules/reports/post-process-report.ts
```diff
@@ -0,0 +1,268 @@
...
+        if (reportData.directionId === firstStationOnLine || reportData.directionId === lastStationOnLine) {
+            return reportData
+        }
+
+        const stationIndex = stationsOnLine.indexOf(reportData.stationId)
+        const directionIndex = stationsOnLine.indexOf(reportData.directionId)
+
+        if (stationIndex === -1 || directionIndex === -1) return reportData
+        if (stationIndex === directionIndex) return reportData
+
```
brandesdavid, line 152: Maybe add a comment over the function sig for faster reading

brandesdavid (approved):
Please help me understand my comments. otherwise apporved

packages/hono-backend/src/common/utils.ts
johan-t: Since we are already possibly returning undefined, typescript will force us to handle the undefined case when calling the function. Otherwise we would make the function impure, since logging or throwing an error introduces a side effect. So I think for a util function like this it is better to handle it on the outside

PR #448, opened 2025-12-22
feat: check some secret info with `verifyRequest`

johan-t:
## Describe the issue:
Previously any person could automatically send POST report requests and thus would have spammed the database.

Before the request is processed we are now sending a request with the headers to a secret microservice, which will decide if the request is valid or not. 

I also added a test case to test the X-Password exception. I am hesitant to add any other test cases since we could accidentally leak any implementation details of the secret microservice.


brandesdavid (approved):
LGTM

brandesdavid (approved):
LGTM, fix conflicts please

PR #449, opened 2025-12-22
feat: server side cache for transit data

johan-t:
## Describe the issue:
For every request to `/transit/lines` or `/transit/stations` we would do a database request. This is useless as the database content would only change when we restart the backend anyways.

Instead we now keep the results of the first database query in memory and simply keep on returning this data.


brandesdavid (approved):
LGTM

PR #450, opened 2026-01-08
feat: predicted reports 

johan-t:
## Describe the issue:
Sometimes we don't have enough report coverage to inform users. In these cases, we predict where inspectors could be located to display at least some data.

I reused the existing guessStation function, which we already use to predict the station when a report is submitted without a stationId. We now calculate a threshold—if current reports fall below it, we use guessStation to predict stations within a given timeframe. If no predictions are found, we expand the timeframe iteratively.
I also added extensive tests for this functionality.

We now call them "predicted reports" instead of "historic reports" to reduce confusion about terminology. This breaks the API contract with the mobile and web apps.


brandesdavid (approved):
LGTM, would recommend to look into codex's comment.

PR #453, opened 2026-02-11
feat: post users feedback

johan-t:
## Describe the issue:
We need a way to accept users feedback that is coming from the web app.

I added a new `/feedback` endpoint which accepts a `POST` request with a `feedback: {...}` body. It returns a `201` when successful. 
We store: feedback, timestamp and user agent.

I added a new migration for the new table. For the other new tables we just modified the default migration for now. For this I felt like a migration is appropriate, let me know if I should include it in the default one.


johan-t:
I have no idea why CI is failing

PR #454, opened 2026-02-11
style: suppress logs tests

johan-t:
I was getting annoyed, since the test output would be spammed with all of the logs that are coming from successful HTTP requests.

Instead we are now suppressing the INFO logs and giving the user the option to enable it again when running the tests


brandesdavid (approved):
lgtm

PR #455, opened 2026-02-16
feat: `GET /reports/:stationId`

johan-t:
## Describe the issue:
When a user clicks on a station they will see all of the reports for this station. In order to show them efficiently we need an endpoint that can return them filtered for a station. 

Added a `GET /reports/:stationId` endpoint which returns only stations for this endpoint.

### DIFF
--- a/packages/hono-backend/src/modules/reports/reports-routes.ts
+++ b/packages/hono-backend/src/modules/reports/reports-routes.ts
@@ -9,43 +9,75 @@ import { insertReportSchema } from '../../db'
 
 import { getDefaultReportsRange, MAX_REPORTS_TIMEFRAME } from './constants'
 
+const reportsQuerySchema = z
+    .object({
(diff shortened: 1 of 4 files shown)


packages/hono-backend/src/modules/reports/reports-routes.ts
```diff
@@ -9,43 +9,75 @@ import { insertReportSchema } from '../../db'
...
+    now = DateTime.now()
+) => {
+    const defaultRange = getDefaultReportsRange(now)
+
+    return {
+        from: query.from ?? defaultRange.from,
+        to: query.to ?? defaultRange.to,
+    }
+}
+
```
mclrc: couldn't we do this in the parser with `default`? if not, i'd still prefer to infer the argument type from the schema here.

mclrc (approved):

PR #456, opened 2026-02-16
docs: document API endpoints

johan-t:
Added documentation that will be auto created and available via an interactive site. This will make it easy to test endpoints.


brandesdavid (approved):
LGTM

PR #457, opened 2026-02-18
feat: `/v0/transit/segments` endpoint

johan-t:
## Describe the issue:
The clients consume the segments to display the lines of the map. We need a way to return them to the users. 

I added a new schema for segments:
```
export const segments = pgTable('segments', {
    id: serial().primaryKey(),
    lineId: varchar({ length: 16 })
        .notNull()
        .references(() => lines.id, { onDelete: 'cascade' }),
    fromStationId: varchar({ length: 16 })
        .notNull()
        .references(() => stations.id, { onDelete: 'cascade' }),
    toStationId: varchar({ length: 16 })
        .notNull()
        .references(() => stations.id, { onDelete: 'cascade' }),
    color: varchar({ length: 7 }).notNull(),
    coordinates: jsonb().notNull().$type<[number, number][]>(),
})
```
Then seed it in `seed.ts` (For now we are just seeding it from the `segments.json`. In the future this will be replaced by the proper seed script where we can recreate them on the fly based on OSM data, making the backend truly stateless).
I also adjusted the default migration to include the new segments table and created an endpoint `/v0/transit/segments` that serializes the data to proper Geojson  and returns it to the client. Since the data is being cached we are only doing this once.

The next step is the ETag caching so that the client avoids requesting this data again and again.

- You have to wipe, migrate and reseed your db if you want to test this PR. 
- The type of one `SegmentFeature` is: `LineString`:
```
type SegmentFeature = {
    type: 'Feature'
    properties: SegmentProperties
    geometry: {
        type: 'LineString'
        coordinates: SegmentRow['coordinates']
    }
}
```
We might have to adjust this, once we write a proper seed db script since some of the lines have divergent paths making them `MultiLineStrings`

### DIFF
--- a/packages/hono-backend/README.md
+++ b/packages/hono-backend/README.md
@@ -1,5 +1,15 @@
 # New Freifahren backend using Hono + Drizzle
 
+## .env
+
+The .env file contains the following variables:
+
+- `DATABASE_URL`: The URL of the database to use.
+- `DATABASE_URL_TEST`: The URL of the test database to use.
+- `NLP_SERVICE_URL`: The URL of the NLP service to use.
+- `REPORT_PASSWORD`: The password to use for the report API.
+- `SECURITY_MICROSERVICE_URL`: The URL of the security microservice to use.
+
 ## Start containers
 
 You can start the DB and bun container like so:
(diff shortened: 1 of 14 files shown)


brandesdavid (approved):
LGTM

PR #458, opened 2026-02-19
feat: add etag caching

johan-t:
## Describe the issue:
When the user opens the app, we need to request the transit data (station, lines and segments). This data can be large (around 1 MB). It would be inefficient to force clients to request this data on every page load even if it did not change.

I added etag caching. This means for the first request we respond with the data, as well as an `ETag` header that contains the hash of the response.
The client can accept this header and for each new request attach the hash using the `If-None-Match:` header. The server compares this header to its computed hash value, if it matches it returns a `304 Not Modified` with no body, else it will return the data. 
This means the client only needs to download the transit data once and store it in localstorage.

I did not add any tests since we are using hono provided middleware. So I think it is safe to trust that hono works.

### DIFF
--- a/packages/hono-backend/README.md
+++ b/packages/hono-backend/README.md
@@ -1,5 +1,15 @@
 # New Freifahren backend using Hono + Drizzle
 
+## .env
+
+The .env file contains the following variables:
+
+- `DATABASE_URL`: The URL of the database to use.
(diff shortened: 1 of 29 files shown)


brandesdavid (approved):
lgtm

PR #459, opened 2026-02-19
feat: cache `GET /v0/reports`

johan-t:
## Describe the issue:
We are requesting the reports every 30 sec on the client. For every request we would touch the database, serialize the data and return it to the client, even if the data did not change since the latest request.
This can create an issue, since the centralized has to handle a lot more transactions then needed, which can limit scalability.

We are now storing the `private latestReportTimestamp: DateTime | undefined = undefined`. It will be updated on every `POST /v0/reports` request, based on the databases response.

For every `GET /v0/reports` request we are returning a `modified-since` header which contains the `latestReportTimestamp`. The client can use this timestamp to include a `if-modified-since` header for subsequent requests, the server compares `latestReportTimestamp` against the `if-modified-since` header, if the servers `latestReportTimestamp` is newer it will return `200` with the reports. If they match the server will return `304 not modified`. 

Since we are storing `latestReportTimestamp` server side, we now only have to request the database if there is actually new content. 

If the `if-modified-since` header that the client sends is older then 5 min, we are returning the data even if the header is still valid. I made this decision since otherwise if there are no new reports for a while, the existing data that is outside of the requested time window would not fade out.
For example:
- at 12:00 a client requests data for the past hour.
- the server returns with a report that is dated at 11:01 and at 11:59
- 2 min later at 12:02 the client requests data again, since no new report has been made in that time the client thinks the data is still valid. Which its not since the 11:01 report is now too old to be displayed.
- The client requests again at 12:05, here the server returns the entire data again even though there has not been a new report, just to inform the client that the 11:01 report is not valid anymore. 

So at the worst case we are accepting data that is 5 min stale. I know this is a little bit hacky but I could not find a better solution. If we run into issues with the database being too busy, I would suggest we track this caching mechanism and see how often we are hitting the cache/ missing it and can decide on a more stable caching mechanism based on this.

### DIFF
--- a/packages/hono-backend/src/index.ts
+++ b/packages/hono-backend/src/index.ts
@@ -11,40 +11,6 @@ import { getReports, postReport, ReportsService } from './modules/reports/'
 import { TransitNetworkDataService } from './modules/transit/transit-network-data-service'
 import { getLines, getStations } from './modules/transit/transit-routes'
 
-const app = new Hono<Env>()
-
-app.use(requestId())
-app.use(
-    pinoLogger({
-        pino: pino({
-            level: process.env.LOG_LEVEL ?? 'info',
-            transport: {
-                targets: [
-                    {
-                        target: 'pino-pretty',
-                        options: {
-                            colorize: true,
-                            ignore: 'pid,hostname,req,res,responseTime,reqId',
-                            translateTime: 'SYS:standard',
-                            destination: 1,
-                        },
-                    },
-                    {
-                        target: 'pino-roll',
-                        options: {
-                            file: './app.log',
-                            frequency: 'daily',
-                            mkdir: true,
-                        },
-                    },
-                ],
-            },
-        }),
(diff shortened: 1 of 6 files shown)


johan-t:
@codex review again

johan-t:
This feature is adding a lot more complexity than I would like to have for now. I would suggest that instead of adding caching out of the box, we setup grafana (selfhost on coolify) and track the performance of this endpoint, if we notice it needs caching we can add it later. Is that fine for you @mclrc @brandesdavid ?

mclrc:
i think this is worth caching, and the extra complexity seems fine, unless there are some issues you discovered that would break things? i think fading old data out could be done client-side. we could explicitly handle the 304 case in the react query/api handler and filter out the older reports from the cached ones, for example.

johan-t:
@codex review again this pr also consider the new approach

PR #460, opened 2026-02-22
refactor: replace `createReportWithTimestamp`

johan-t:
## Describe the issue:
We were using the `createReportWithTimestamp` function, which would allow us to add reports in tests at a later time. The issue with this was that we would add them directly to the database since the API does not accept a timestamp parameter. This means that we would essentially mock the POST reports API endpoint instead of actually using it, moving the test environment a little out of touch with the actual production environment.

Instead we are now just using `Settings.now` to set the local time of the system, this means we just simulate the wait time and submit the report via the API. This makes the tests a little more realistic and thus more reliable. 

## Notes:
We are now passing the local time from the server to the database for the `timestamp` creation of the report, instead of relying on the timestamp support by postgres. This is necessary so that the `Settings.now` work.

### DIFF
--- a/packages/hono-backend/src/modules/reports/reports-service.ts
+++ b/packages/hono-backend/src/modules/reports/reports-service.ts
@@ -250,7 +250,7 @@ export class ReportsService {
             timestamp: Date
         }
     }> {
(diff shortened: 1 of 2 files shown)


brandesdavid (approved):
lgtm

PR #461, opened 2026-02-22
feat: run the risk model

johan-t:
I moved the risk model to the new backend, it now works in a similar way we are still running it as a python sub process, now with uv to make handling dependencies a little easier (I also adjusted the dockerfile to pull uv on start up). 
We are now also passing in the report AND segments with `stdin` instead of relying on a `segments.json` file.

I added extensive tests for this feature. 

I had to make some changes in the tsconfig to handle the sub process

Next step is caching

### DIFF
--- a/packages/hono-backend/src/modules/risk/risk-service.ts
+++ b/packages/hono-backend/src/modules/risk/risk-service.ts
@@ -1,0 +1,60 @@
+import { join } from 'path'
+
+import { DateTime } from 'luxon'
+
+import { AppError } from '../../common/errors'
+import type { ReportsService } from '../reports/reports-service'
+import type { TransitNetworkDataService } from '../transit/transit-network-data-service'
+
+type RiskData = {
+    segments_risk: Record<string, { color: string; risk: number }>
+}
+
+export class RiskService {
+    constructor(
+        private reportsService: ReportsService,
+        private transitNetworkDataService: TransitNetworkDataService
+    ) {}
+
+    async getRisk(): Promise<RiskData> {
+        const now = DateTime.utc()
+        const oneHourAgo = now.minus({ hours: 1 })
+
+        const [segmentsCollection, realReports, stations] = await Promise.all([
+            this.transitNetworkDataService.getSegments(),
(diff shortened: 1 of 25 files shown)


packages/hono-backend/src/modules/risk/risk-service.ts
```diff
@@ -0,0 +1,60 @@
...
+        proc.stdin.write(JSON.stringify({ segments: segmentsCollection.features, reports: pythonReports }))
+        await proc.stdin.end()
+
+        const exitCode = await proc.exited
+        if (exitCode !== 0) {
+            throw new AppError({
+                message: 'Risk model failed',
+                statusCode: 500,
+                internalCode: 'RISK_MODEL_FAILED',
+                description: await new Response(proc.stderr).text(),
```
mclrc, line 54: condition on dev env?

mclrc (approved):
i assume the risk model script is more or less the same? if so, good to go in my opinion. we should add tests for this in the future but for now it should be fine if the core logic is the same.

packages/hono-backend/src/modules/risk/risk-service.ts
johan-t: The error handler, strips the description on production before returning to the client. That way we log it but do not leak anything:

```
export const handleError = (err: Error, c: Context) => {
    if (err instanceof AppError) {
        c.get('logger').error(
            {
                internal_code: err.internalCode,
                statusCode: err.statusCode,
                description: err.description,
            },
            err.message
        )
        return c.json(
            {
                message: err.message,
                details: {
                    internal_code: err.internalCode,
                    // We do not want to leak sensitive information to the client in production
                    description: process.env.NODE_ENV === 'development' ? err.description : undefined,
                },
            },
            err.statusCode
        )
    }

    c.get('logger').error(err, 'Unhandled error')
    return c.json(
        {
            message: 'Internal Server Error',
            details: {
                internal_code: 'UNKNOWN_ERROR',
                // We do not want to leak sensitive information to the client in production
                description: process.env.NODE_ENV === 'development' ? err.message : undefined,
            },
        },
        500 as ContentfulStatusCode
    )
}
```

PR #463, opened 2026-03-10
Disable link preview options in send_message

### LINKED ISSUE #442: Disable link preview for lucidity
Frequent messages from the Telegram bot make the group chat very crowded. Please consider disabling the link preview to make the messages smaller and enable users to see more messages without scolling.

dierigoletto:
Fixes #442

The bot currently sends messages with `prefer_large_media=True`, which renders 
the FreiFahren logo as a large image on every inspector report. With hundreds 
of messages per day, this creates significant visual noise in the group chat.

This fix disables the link preview directly in the bot's `send_message` function 
in `telegram_adapter.py`.

As an alternative `prefer_small_media=True` is also available. This might mititage some of the noise.

### DIFF
--- a/packages/nlp_service/services/telegram_adapter.py
+++ b/packages/nlp_service/services/telegram_adapter.py
@@ -46,8 +46,7 @@ def send_message(
         parse_mode (str): Message parsing mode (default: HTML).
     """
     try:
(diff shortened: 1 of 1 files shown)


johan-t:
The new version of the bot added this feature, thanks for the suggestion :)

PR #471, opened 2026-05-07
refactor: adjust frontend for new Hono backend

johan-t:
## Summary

This PR adapts the web frontend to the new Hono backend transit/reporting API shape.

The main user-facing change is that the frontend now works with backend-provided transit entities directly: stations, line variants, segments, risk data, and reports are all consumed from the new `/v0` endpoints. The report, map, and reports modal flows were updated to use the new flat report shape and the new transit network data model.

### New backend API contract

The frontend now uses the new Hono backend routes for core data:

- reports are loaded from `/v0/reports`
- station-scoped reports are loaded from `/v0/reports/:stationId`
- stations are loaded from `/v0/transit/stations`
- lines are loaded from `/v0/transit/lines`
- segments are loaded from `/v0/transit/segments`
- risk is loaded from `/v0/risk`
- distance is loaded through the new `from` / `to` query parameters

Reports are now treated as flat payloads with IDs:

- `stationId`
- `lineId`
- `directionId`
- `timestamp`
- `isPredicted`

The frontend resolves station and line display names from the transit datasets instead of expecting nested station/line objects in each report.

### Report creation flow

The report form now submits reports using the new backend payload:

- `stationId`
- `lineId`
- `directionId`
- `source: "web_app"`

The old free-text report message flow was removed from the web report form because the new report endpoint does not accept that shape.

Validation was adjusted to work from station coordinates directly instead of the old nested report object.

### Split-line handling

The backend may expose split line variants such as `M1-a` and `M1-b`, while users should still see the public line name `M1`.

For now, the web report form handles this locally:

- line variants are grouped by their public display name
- the user sees one public line option, e.g. `M1`
- stations from all variants are shown for that public line
- direction options are built from the variants that contain the selected station
- when a direction identifies a concrete variant, that variant is submitted
- when the report is still ambiguous, the longest matching variant is submitted as a temporary fallback

This is intentionally a temporary frontend-side compromise until the backend supports multiple candidate line IDs for ambiguous reports.

### Public line display

User-facing surfaces now resolve backend line variant IDs to public line names.

Examples:

- `M1-a` and `M1-b` are displayed as `M1`
- station info line badges are deduplicated by public line name
- report modal line charts merge split variants into one public line row
- current risk entries aggregate by public line name
- shared line badges and share text use the resolved public line name

This keeps backend IDs available internally while avoiding variant IDs leaking into the UI.

### Map and risk rendering

Map segment rendering was updated for the new segment shape.

Regular lines now use the segment `color` property from the backend. Risk overlay rendering now matches risk entries by numeric segment ID, which is the shape returned by the new risk endpoint.

The Reports Modal current-risk summary was also fixed to map risk segment IDs back through the segment collection before aggregating by line. This prevents numeric segment IDs from appearing as line labels.

Station map clicks were made more defensive. The click handler now checks which station layers are actually present before querying rendered features, preventing crashes when a layer is not yet available in the MapLibre style.

### Reports modal and marker updates

The reports modal, latest-report button, report markers, marker detail modal, station info modal, share button, and viewed-report tracking were updated to work with the new flat report structure.

Predicted reports now replace the previous historic-report naming in the frontend state and display logic.

### Navigation removal

The old web navigation modal and related route UI were removed from this frontend path. The new backend adaptation focuses on reporting, stations, lines, segments, map rendering, and risk display.

### Backend/report model follow-up

There is still a known modeling gap around ambiguous split-line reports.

The desired future backend behavior is for report creation to accept:

```ts
lineIds: string[] | null
```

This would allow the frontend to submit multiple candidate variants for ambiguous public-line reports instead of choosing the longest variant. Backend post-processing should then narrow candidates using `directionId` when possible.

https://github.com/FreiFahren/FreiFahren/issues/481


johan-t:
This can only be merged when the new backend is deployed and the mobile app update is ready to be released.

brandesdavid (approved):
LGTM

PR #497, opened 2026-05-20
test: add Telegram bot integration tests against real Hono backend

johan-t:
## Summary

Closes FRE-591. Adds an integration test suite for the Telegram bot that runs against the **real Hono backend** (started via the existing `compose-hono.test.yaml` stack) so backend contract drift — schema changes on `/v0/transit/*`, `POST /v0/reports`, path versioning, the password header — fails the bot's tests instead of production.

- **35 tests** covering inbound (`Telegram message → /v0/reports`), outbound forwarding (`backend → bot /report → Telegram`), and the spam classifier.
- Mistral and the outbound Telegram API are the only stubs. Everything else (the bot's handler, the aiohttp forwarding server, `httpx` calls to the backend) is exercised end-to-end.
- One-line refactor in `main.py`: `handle_text` lifted from a closure in `run()` to the module-level factory `build_handle_text(...)` so tests construct it with the same code path as prod.
- CI consolidated: the previous `run-python-bot-tests.yml` (stale Python 3.9, pointed at the obsolete `nlp_service`) is deleted; the bot tests now run as additional steps in the renamed `run-backend-ci.yml`, reusing the compose stack so we don't pay for a second migrate+seed.

## Test plan

- [x] Local: `docker compose -f packages/hono-backend/compose-hono.test.yaml up -d --wait && ... db:migrate && ... db:seed && cd packages/telegram_bot && uv run pytest` → 35 passed
- [x] `uv run ruff check .` clean
- [x] CI green on this PR (verifies the merged workflow brings up the stack, runs hono + bot tests, and tears down)
- [x] Sanity-check that if `/v0/transit/lines` shape changes, the `transit` session fixture fails fast and every bot test errors at setup

## Notes

- Local docs: a `## Test` section in `packages/telegram_bot/README.md` documents the three-command flow (compose up → migrate+seed → pytest).
- DB isolation: hono tests connect to `freifahren_test` (via `DATABASE_URL_TEST` in `preload.ts`); the bot tests hit the bun service at `freifahren`. They don't share a DB, so the order in the workflow is irrelevant.


johan-t:
@codex review this PR

PR #520, opened 2026-05-25
feat: implement contribution modal and update environment configuration

brandesdavid:
- Add a new ContributionModal component for user contributions, including options for Stripe and bank transfers.
- Update Dockerfile to use Node 20 and pass environment variables for Stripe configuration.
- Create .dockerignore to exclude unnecessary files from the Docker build context.
- Introduce .env.example for environment variable documentation.
- Modify compose.yaml to include new build arguments for frontend service.
- Enhance localization files (en.json, de.json) with contribution-related text.
- Add a copy icon SVG for the contribution modal.
- Update UtilModal to include a button for opening the ContributionModal.

## Summary

This PR adds a **Contribution** popup to the legacy web app (`packages/frontend`) so users can support FreiFahren via Stripe or bank transfer.

- **`ContributionModal`** with two tabs: Stripe Buy Button (via `<stripe-buy-button>`) and bank details (holder, IBAN, BIC, reference) with per-field copy icons (`/icons/copy.svg`, checkmark feedback via `risk-0.svg`).
- **When it appears:** once per session after a 3 min timer on load, after closing the report summary modal, or manually from the Util modal (“Beitrag leisten” / “Contribute”) — the Util path bypasses the “don’t show again” preference.
- **Dismissal:** “Nicht wieder anzeigen” persists `contributionModalDismissed` in `localStorage`.
- **Stripe config** via `VITE_STRIPE_BUY_BUTTON_ID` and `VITE_STRIPE_PUBLISHABLE_KEY` (documented in `.env.example`; no secret keys in the frontend).
- **Docker/build fixes:** `.dockerignore` (fixes esbuild/`node_modules` clash), Node 20, correct `dist` output path, build args for `VITE_API_URL` and Stripe vars in `compose.yaml`.

## Test plan

- [ ] Copy `packages/frontend/.env.example` → `.env`, set Stripe vars, run `npm run dev` — Stripe tab renders the buy button.
- [ ] Bank tab: copy each field; icon switches to checkmark briefly.
- [ ] Wait 3min on first visit — modal opens; close with X vs “Nicht wieder anzeigen” — reload verifies permanent dismiss only for the latter.
- [ ] Submit a report → close summary → contribution modal opens (if not dismissed).
- [ ] Util modal → “Beitrag leisten” opens contribution modal even when dismissed via localStorage.
- [ ] `docker compose build frontend` with `VITE_STRIPE_*` set (root `.env` or exports) — app loads on `:3000`, API calls hit `VITE_API_URL`, Stripe tab works in the built image.
- [ ] Close button sits fully above the modal (not clipped).

## Notes for reviewers

- Publishable Stripe keys are intentionally client-side; only `pk_*` and buy button ID belong in `VITE_*` env vars.
- Timed trigger is currently **5 seconds** (likely for local testing) — confirm whether production should use a longer delay (e.g. 3 minutes) before merge.
- Bank IBAN/BIC are hardcoded in the component; update there or move to config if they change.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -63,9 +63,11 @@ services:
     build:
       context: ./packages/frontend
       dockerfile: Dockerfile.frontend
+      args:
+        VITE_API_URL: http://localhost:8080
+        VITE_STRIPE_BUY_BUTTON_ID: ${VITE_STRIPE_BUY_BUTTON_ID:-}
+        VITE_STRIPE_PUBLISHABLE_KEY: ${VITE_STRIPE_PUBLISHABLE_KEY:-}
     container_name: frontend
-    env_file:
-      - ./packages/frontend/.env
     ports:
       - "3000:3000"
     depends_on:
--- a/packages/frontend/.dockerignore
+++ b/packages/frontend/.dockerignore
@@ -1,0 +1,9 @@
+node_modules
+dist
+build
+.git
+.env
+.env.*
+*.log
+.DS_Store
+coverage
--- a/packages/frontend/.env.example
+++ b/packages/frontend/.env.example
@@ -1,0 +1,6 @@
+VITE_JAWG_ACCESS_TOKEN=
+VITE_API_URL=http://localhost:8080
+VITE_STRIPE_BUY_BUTTON_ID=buy_btn_xxx
+VITE_STRIPE_PUBLISHABLE_KEY=pk_live_xxx
+SENTRY_AUTH_TOKEN=
+NAVIGATION_SERVICE_URL=https://api.transitous.org/api/v1/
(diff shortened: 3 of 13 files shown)


johan-t (approved):
The language is a bit off, here everything is in english expect for the "Unterstützen"
Image

I would maybe also mention for what we need the money e.g. one small sentence to mention that all money is being reinvested into the app

I think we can really optimize this by:
1. making it prettier (we can maybe take a look at some other examples)
2. doing a cost breakdown page where we show what the money is being spent on
3. doing multiple tiers so that the user does not have to choose the amount
4. doing recurring payments

but i think this is good enough and will be rewritten in a couple of weeks/ days anyway

PR #525, opened 2026-05-28
fix: remap reports from orphan stations during reseed

johan-t:
## Summary

Fixes FRE-609 by adding a conservative reseed repair pass before orphan station pruning.

- Remap report `station_id` and `direction_id` references from line-less orphan stations to a single clear live replacement.
- Reuse seed station name/distance matching rules and require line-serving candidates when reports carry a `line_id`.
- Log unresolved or ambiguous orphan references, then delete reports that still point at orphan stations so pruning can remove stale station rows.
- Add integration-style seed tests that create reports through the API.

## Validation

- `./node_modules/.bin/eslint src/db/seed/seed.ts --ext .ts`
- `bunx tsc --noEmit --skipLibCheck`
- Previously ran full backend suite against local Postgres: `91 pass, 0 fail`


johan-t:
@codex review

PR #602, opened 2026-06-07
fix: stop re-asking location with the in-app prompt on every iOS visit

johan-t:
## The bug

On iOS Safari, after allowing location once via the in-app soft-ask, returning to the page showed the **in-app prompt again on every visit**, while the **native popup never reappeared** and the per-site setting stayed at "Fragen" (Ask).

## Root cause — and the part that is iOS, not us

**iOS Safari scopes a geolocation grant to the session.** Tapping "Erlauben" in the native prompt grants location for that session only; it does **not** flip the per-site setting to "Allow", and `navigator.permissions.query({name:'geolocation'})` keeps returning `'prompt'` indefinitely (documented WebKit behavior — there is no persistent "remember" on the iOS prompt).

The previous change (#601) made the in-app soft-ask the gate for *every* `'prompt'` result. Since iOS always reports `'prompt'`, users got re-asked on every visit. We cannot detect prior consent through the Permissions API on iOS, so we must remember it ourselves.

## The fix

Distinguish **first-time** from **returning consenter**:

```ts
if (permission === 'denied') return;
if (permission === 'granted' || hasConsentedToLocation()) { trigger(); return; }
// first-time prompt/unsupported → deferred in-app soft-ask
```

- `hasConsentedToLocation()` reads a new `locationConsented` localStorage flag, set **only** when the user explicitly taps **Allow** in our in-app card. This guarantees the soft-ask is shown at least once before we ever skip it — so #601's behavior (first-timers always see the in-app prompt, never an uninvited native one) is preserved.
- A **new** localStorage key is used so stale `locationOptIn` values from the earlier build don't pre-qualify anyone.

### Resulting behavior (iOS Safari "Ask")

| Scenario | Behavior |
|---|---|
| First visit | soft-ask → Allow → native prompt → location |
| Return, same session | triggers directly → silent → dot, **no in-app prompt** |
| Return, new session (grant lapsed) | triggers directly → **browser's native prompt** reappears, **no in-app nag** |

Chrome / Firefox / Android persist the grant normally, so `query()` returns `granted` and they track silently with no prompts at all on return.

## What this does NOT (and can't) fix

The per-site setting staying at "Fragen" and the native prompt reappearing each new session is **inherent iOS Safari behavior** (session-scoped geolocation). The only way to stop all prompting is the user manually setting the site's Location to "Allow" in Safari's website settings (aA menu → Website Settings → Location → Allow). No web app can change that programmatically.

Follow-up to #601 / #598 (FRE-635).

### DIFF
--- a/packages/frontend-next/src/components/map/UserLocationControl.tsx
+++ b/packages/frontend-next/src/components/map/UserLocationControl.tsx
@@ -7,8 +7,10 @@ import { useGeolocation } from '@/contexts/Geolocation.context';
 import { useLegalDisclaimer } from '@/lib/legal-disclaimer';
 import {
   dismissLocationPrompt,
+  hasConsentedToLocation,
   isLocationPromptDismissed,
   LOCATION_PROMPT_DELAY_MS,
+  markLocationConsented,
   queryGeolocationPermission,
 } from '@/lib/location-prompt';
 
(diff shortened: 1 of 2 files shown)


johan-t:
Closing without merging. After researching permission-priming UX best practices, the decision is to **always show the in-app prime before the native prompt — including for returning users** (a context-free native re-ask is granted ~40% less often, per web.dev). That is exactly what main already does (#601): 'granted' tracks silently; every 'prompt'/'unsupported' result shows the in-app soft-ask first. This PR did the opposite (skip the in-app card for returning consenters), so it's obsolete. The redundant in-app card on within-session iOS revisits is the accepted trade-off, since iOS doesn't let us distinguish a live session grant from a lapsed one.

PR #708, opened 2026-06-28
feat: cache transit responses at the edge via the Workers Cache API

johan-t:
## Problem

The dashboard Cache Rule on `/v*/transit/*` never cached anything: Cloudflare Cache Rules only act on a Worker's outbound `fetch()` subrequests to an origin, never on a response the Worker generates itself. Since `api.freifahren.org` routes straight to the api-worker and the transit payloads are built from D1 (no origin fetch), every request recomputed the response (no `cf-cache-status` on responses).

## Change

Cache the `GET /v0/transit/{stations,lines,segments}` responses inside the worker via the **Workers Cache API** (`caches.default`), keyed by URL:

- The cache lives inside the worker, so **every caller benefits** — the browser app and the telegram-worker's server-side fetches alike.
- Entries are stored **origin-agnostic** (CORS headers stripped); `cors()` re-adds the correct per-origin headers on the way out, so the CORS allowlist still holds.
- Entries retain the `Cache-Tag`, so the existing **purge-by-tag** on reseed keeps working (no purge changes).
- `If-None-Match` is honored on a hit, so the **304 revalidation** path the frontend's localStorage cache relies on is preserved.
- Inert under unit tests (no Cache API / execution context in Bun) — falls through to the live handler.

Middleware order is now `cors → transitEdgeCache → transitCacheMiddleware → etag → handler`, so the stored entry carries the ETag and `Cache-Control`/`Cache-Tag` before it's written.

## Verify after deploy

```
curl -sD - -o /dev/null https://api.freifahren.org/v0/transit/segments | grep -i x-edge-cache
```

First request → `X-Edge-Cache: MISS`, second → `HIT`.

## Notes

- The dashboard Cache Rule is now redundant (it can't act on worker-generated responses) and can be deleted.
- 89/89 existing tests pass; lint clean.

### DIFF
--- a/packages/api-worker/src/index.ts
+++ b/packages/api-worker/src/index.ts
@@ -10,6 +10,7 @@ import { getReports, getReportsByStation, postReport } from './modules/reports/'
 import { getRisk } from './modules/risk/risk-routes'
 import {
     transitCacheMiddleware,
+    transitEdgeCacheMiddleware,
     VERSIONED_TRANSIT_CACHEABLE_PATHS,
     VERSIONED_TRANSIT_PATH,
 } from './modules/transit/transit-cache-middleware'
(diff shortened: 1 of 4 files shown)


johan-t:
@codex review

PR #829, opened 2026-07-17
fix: refresh PWA after service-worker update

### LINKED ISSUE #828: fix: refresh the PWA after a service-worker update
## Problem

Installed web-app clients can continue running an old JavaScript bundle after a new service worker is available. A service-worker activation alone does not reload the already-open page.

## Minimal fix

Use `virtual:pwa-register`, already provided by `vite-plugin-pwa`, instead of the handwritten registration.

* Check for an update when the app starts and when it returns online or to the foreground.
* Reload once when the updated worker takes control.
* Keep registration failures non-fatal and avoid unhandled rejections.
* Remove `html` from the precache glob so the HTML shell remains network-first.

Do not add a separate PWA framework or redesign offline caching.

## Acceptance criteria

* An existing installed PWA running an old build switches to the latest build after returning online/foregrounded.
* The page reloads at most once per update.
* Offline use remains available from the last cached shell.
* `bun run lint`, `bun run format:check`, and `bun run build` pass.

johan-t:
## Summary

Fixes the installed web PWA remaining on an old JavaScript bundle after a new service worker is available.

- Replace the handwritten service-worker registration with `virtual:pwa-register`, already supplied by `vite-plugin-pwa`.
- Check for a worker update at registration, then only when connectivity returns or the app returns to the foreground.
- Configure `skipWaiting` and `clientsClaim` explicitly because registration injection is disabled.
- Exclude `index.html` from precache so the HTML release pointer remains network-first.
- Warm the runtime `app-shell` cache with `/` on first online registration, preserving the first offline cold start.
- Keep the PWA plugin resolvable but disabled for Capacitor builds, so native continues to emit no service worker.
- Add browser-level PWA regression coverage through one frontend command: `bun run test`.

Closes #828.

## Root cause

The custom registration intentionally swallowed service-worker registration failures, but it did not manage the update lifecycle. An installed PWA could therefore keep executing the already-open bundle after a deployment.

With `injectRegister: false`, `registerType: autoUpdate` also does not implicitly configure the generated worker to skip waiting and claim existing clients. The updated worker needs explicit activation settings and page-side lifecycle handling.

A first installation needs a runtime-shell warm-up: the install navigation happens before the worker controls the page, so it cannot populate the `NetworkFirst` cache itself.

## Resulting update path

An online installed PWA checks `/sw.js` at registration and warms a missing `/` entry in `app-shell`. When it reconnects or returns to the foreground, it rechecks. A changed worker activates immediately, takes control of existing clients, and vite-plugin-pwa reloads the page once so it fetches the current network-first HTML shell and matching hashed assets.

Offline behavior remains available for returning users, including a newly installed app after its first online registration. This PR does not add another PWA framework or redesign offline caching.

## Browser regression suite

`bun run test` runs the PWA suite in Chromium. It builds the real frontend, starts the real local API Worker in `workerd`, and seeds Berlin data into an isolated temporary local D1 state. The test verifies:

1. The browser can query real seeded transit data through the API and its CORS policy.
2. The installed service worker warms the `app-shell` cache.
3. A fresh offline page load renders from that cached shell.
4. After connectivity returns, navigation obtains a newer network response instead of reusing the cached shell.

The static frontend server is intentionally only a local host for the generated PWA assets and the changing-shell assertion. It does not mock API endpoints. The temporary Worker persistence directory is removed after the run, so future browser journeys can set up state through API calls without modifying a developer local D1 or production data.

## Validation

- `bun run test`
- `bun run format:check`
- `bun run lint` (passes with the two existing TanStack Virtual / React Compiler warnings)
- `CAPACITOR=true bun run build`
- `bun run build`
- Inspected the generated web service worker: it includes `skipWaiting` and `clientsClaim`, and does not precache `index.html`.
- Verified the Capacitor build emits neither a service worker nor a web manifest.

### DIFF
--- a/.github/workflows/frontend-next-ci.yml
+++ b/.github/workflows/frontend-next-ci.yml
@@ -28,6 +28,9 @@ jobs:
             - name: Install dependencies
               run: bun install --frozen-lockfile
 
+            - name: Install API test dependencies
+              run: bun --cwd ../api-worker install --frozen-lockfile
+
             - name: Run Prettier check
               run: bun run format:check
 
@@ -38,3 +41,9 @@ jobs:
             # build failures — neither of which the lint/format steps above catch.
             - name: Run build
               run: bun run build
+
+            - name: Install Chromium
+              run: bunx playwright install --with-deps chromium
+
+            - name: Run PWA integration test
+              run: bun run test
--- a/.github/workflows/frontend-next-ci.yml
+++ b/.github/workflows/frontend-next-ci.yml
@@ -29,7 +29,7 @@ jobs:
               run: bun install --frozen-lockfile
 
             - name: Install API test dependencies
-              run: bun --cwd ../api-worker install --frozen-lockfile
+              run: bun install --cwd ../api-worker --frozen-lockfile
 
             - name: Run Prettier check
               run: bun run format:check
--- a/packages/api-worker/src/db/seed/seed-d1.ts
+++ b/packages/api-worker/src/db/seed/seed-d1.ts
@@ -35,6 +35,15 @@ const toSqlLiteral = (value: unknown): string => {
 // Invoke the locally-installed wrangler CLI (via npx so the package's own binary is used).
 const wrangler = (...args: string[]) => execFileSync('npx', ['wrangler', ...args], { stdio: 'inherit' })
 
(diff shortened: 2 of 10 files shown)


johan-t:
@codex review

johan-t:
@codex review

PR #832, opened 2026-07-18
refactor: make D1 migration commands city-aware

johan-t:
## Describe the issue:

Closes [FRE-743](https://linear.app/freifahren/issue/FRE-743/make-d1-migration-commands-city-aware). Follow-up to FRE-714 (#817).

Each city runs on its own isolated D1 database (Berlin → `DB`, Leipzig → `DB_LEIPZIG`), but the migration commands were hard-coded to Berlin's binding:

```jsonc
"db:migrate":        "wrangler d1 migrations apply DB --local",
"db:migrate:remote": "wrangler d1 migrations apply DB --remote",
```

That left three problems:

1. **No supported way to migrate any city but Berlin.** Every other city's schema had to be applied out-of-band with a hand-written binding — which is why FRE-714 needed a dedicated Leipzig migration job as a stopgap.
2. **Duplicated `wrangler d1 migrations apply` logic** across the `db:migrate` scripts, the seed CLI, and the CI deploy fan-out — free to drift apart. The seed's copy became redundant once Leipzig joined the seed registry.
3. **Binding resolution wasn't tied to the registry**, so the commands couldn't stay correct automatically as cities are added.

The migrate commands now resolve each city's binding from the shared `@freifahren/cities` registry — no binding is hard-coded — and one shared helper backs every caller.

- **Registry accessors** (`packages/cities`): added `getCityDatabase(slug)`, `CITY_DATABASE_SLUGS`, and `CityDatabaseSlug` over the existing `CITY_DATABASES`. Deliberately backed by the provisioned-database registry (not the runtime city registry), so a database can be migrated before its transit config lands.
- **Shared helper** `src/db/migrate.ts` — `applyMigrations({ binding, remote, persistTo? })`, a single wrapper around `wrangler d1 migrations apply`. Idempotent via wrangler's `d1_migrations` ledger.
- **City-aware CLI** `src/db/migrate-d1.ts`, wired to `db:migrate` / `db:migrate:remote`:
  - with **no `--city`**, fans out over **every** provisioned city so all databases stay on one schema;
  - with `--city <slug>`, targets a single city;
  - an unknown slug fails fast before touching any database.
- **De-duplication**: the seed CLI (`seed-d1.ts`) and the CI deploy job now migrate through the same helper/command instead of owning their own `wrangler` invocations. The deploy job applies with one `db:migrate:remote` and keeps the per-database drift guard that reads each `d1_migrations` ledger and fails if the databases land on different heads.
- **Docs**: the API README documents the city-aware workflow (all cities by default, `--city` to target one).

Verified end-to-end locally:

- From a clean `.wrangler/state`, `bun run db:migrate` applied migrations to both `DB` (berlin) and `DB_LEIPZIG` (leipzig); re-running is a no-op (`✅ No migrations to apply!`).
- `bun run db:migrate --city leipzig` touched only `DB_LEIPZIG`; `--city atlantis` exits 1 with a clear error.
- `bun run seed --city berlin` completes through the shared helper.
- `bun run test` → **144 passed / 13 files**; lint, prettier, and typecheck clean.

Schema state is intentionally kept identical across all cities for now — the fan-out default enforces that, and the deploy drift guard fails the deploy if they ever diverge.

@codex review

---
_Generated by [Claude Code](https://claude.ai/code/session_01CKiEvjqRv3Dh3p9RoaeyLV)_


johan-t:
@codex review

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #833, opened 2026-07-18
feat: add lazy station insights endpoint

johan-t:
## Summary

Introduces `GET /v0/insights/station/:stationId` as the historical-data contract for the upcoming station sheet.

The endpoint deliberately separates durable historical insights from live activity:
- returns an exported Zod contract with a full local-week profile (7 × 24 buckets), resolved scope, a 30-day station report count, and rank metadata;
- resolves profile scope as station (≥250 all-history reports), then only a single served line (≥1,500 reports), otherwise city;
- derives local buckets using each city’s IANA timezone, so DST is handled outside SQL;
- sets a 24-hour Workers edge TTL while browsers revalidate on every use.

It does not add KV, cron, scheduled work, or a live-activity endpoint. Clients derive last-hour/today counts from the existing reports API and exclude predicted reports.

## Why

The original rollup/KV proposal optimized aggregation before it had a consumer. The Worker already has a lazy CDN-cache path: D1 computes an insight response only on an edge miss, then Workers Cache serves it for 24 hours. This keeps the public contract stable while avoiding storage and scheduling infrastructure.

## Validation

- `bunx vitest run tests/getInsights.test.ts`
- focused Prettier and ESLint checks
- `bunx wrangler deploy --dry-run`
- read-only production D1 checks for station, single-line, and multi-line fallback cases

The full API-worker suite also runs, with three pre-existing failures in randomized predicted-report threshold assertions; they reproduce when `tests/getReports.test.ts` runs alone.

### DIFF
--- a/packages/api-worker/src/app-env.ts
+++ b/packages/api-worker/src/app-env.ts
@@ -5,6 +5,7 @@ import { Context, Hono } from 'hono'
 import { AppError } from './common/errors'
 import { createLogger, Logger, LogLevel } from './common/logger'
 import { createD1Db, DbConnection } from './db'
+import { InsightsService } from './modules/insights'
 import { ReportsService } from './modules/reports'
 import { RiskService } from './modules/risk'
 import type { CacheCtx } from './modules/transit/reference-cache'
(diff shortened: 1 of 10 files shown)


johan-t:
@codex review

PR #841, opened 2026-07-19
fix: propagate circular-line risk in both directions

### LINKED ISSUE #837: fix: propagate circular-line risk in both directions
## Observed

A recent S42 report at Gesundbrunnen colours the Ringbahn risk layer in only one direction. The other adjacent branch remains green. A circular service has no terminal, so the ordered segment list must wrap when calculating distance from a reported station.

## Investigate

* Reproduce the behaviour against the current risk model using a minimal circular topology.
* Identify the exact topology/distance assumption responsible (including the `stationRank` lookup and spatial decay calculation in `predictSegmentRisk`).
* Confirm linear-line behaviour and directional-report semantics remain correct.

## Fix

Update circular-line propagation so risk reaches the relevant segments on both sides of a reported station and distances around the ring use the correct wrapped/shortest path.

## Regression test

Add a city-agnostic unit test for `predictSegmentRisk` using synthetic station IDs and a closed loop of segments. A report at one loop station must produce non-green risk on both adjacent loop branches; it must not rely on Berlin/S42/Gesundbrunnen seed data.

## Acceptance criteria

* A circular line report propagates through both directions at the reporting station, including the list boundary where the first and last segments meet.
* Linear topology behaviour is unchanged.
* The new city-agnostic regression test fails before the fix and passes after it.
* Verify the repaired behaviour at Gesundbrunnen on the deployed/local risk layer.

johan-t:
## Describe the issue:

A report on a circular line only coloured the risk layer in one direction (FRE-752, #837). Reproduced with a fresh S42 report at Gesundbrunnen: only the Wedding-side branch turned red/yellow while everything from Schönhauser Allee around the ring stayed green.

`predictSegmentRisk` measures a segment's distance from the reported station as the absolute difference between the segment's rank and the station's rank in the line's ordered segment list. Gesundbrunnen is position 0 in S42's list, so segments on the other adjacent branch sit at ranks 23–25, the spatial decay (beta-binomial, zero beyond distance 4–6) evaluated to 0, and that whole side stayed green. A circular service has no terminal, so the ordered segment list must wrap.

- **Wrapped distance for rings** (`risk-model.ts`): for circular lines the distance is now the shorter way around the ring, `min(d, ringSize − d)`. Linear lines keep the unwrapped distance, and directional-report semantics are untouched.
- **Ring size**: seeded rings drop the closing station repeat (the `line_stations` PK forbids it), so their segment list is open by exactly one arc — S41/S42 have 27 stations but 26 segments. The wrap modulus for a flagged circular line is therefore `segment count + 1`. A segment list that already closes on itself (last `toStationId` === first `fromStationId`) wraps with its own length and is treated as a ring even without the flag.
- **Plumbing** (`risk-service.ts`): `RiskService` resolves circular line ids from the (cached) lines table and passes them to `predictSegmentRisk` as a new optional parameter, keeping the model itself city-agnostic.

With the fix, the same Gesundbrunnen report colours both branches symmetrically (verified in the Vitest suite and against a locally seeded `wrangler dev` API):

```
pos 0  Gesundbrunnen → Wedding            1.000 dark red      pos 25 Prenzlauer Allee → Schönhauser  0.831 red
pos 1  Wedding → Westhafen                0.980 dark red      pos 24 Greifswalder → Prenzlauer       0.619 red
pos 2  Westhafen → Beusselstraße          0.831 red           pos 23 Landsberger → Greifswalder      0.357 yellow
pos 3–4 fade to yellow, pos 5–22 stay green (wrap is a shortest path, not a blanket)
```

Tests:
- New city-agnostic unit test for `predictSegmentRisk` (`tests/risk-model.test.ts`) using real station ids and a closed loop of segments — a report at one loop station must produce non-green risk on both adjacent branches, including across the list boundary. It also covers the open-list + flag shape the seed produces, unchanged linear behaviour, and directional reports on a ring. All circular assertions fail on `main` and pass with the fix.
- New integration test in `tests/getRisk.test.ts` that reports at the segment-list boundary station of whatever circular line the seeded city has and asserts both boundary segments come back non-green through the public API.

- The arc between the boundary station and the previous station (e.g. Schönhauser Allee → Gesundbrunnen) belongs only to the sibling ring line (S41), so an S42-only report still leaves that single arc green on S42's layer; risk reaching it would require spreading a report across sibling ring lines, which is report-semantics work outside this fix.
- `buildLineIndexes` was extracted from `predictSegmentRisk` unchanged to keep the function under the ESLint complexity ceiling.

Fixes FRE-752
Closes #837

https://claude.ai/code/session_01MYWRQkkyWDoU5Nho9n35fH

---
_Generated by [Claude Code](https://claude.ai/code/session_01MYWRQkkyWDoU5Nho9n35fH)_

### DIFF
--- a/packages/api-worker/src/modules/risk/risk-model.ts
+++ b/packages/api-worker/src/modules/risk/risk-model.ts
@@ -2,7 +2,8 @@
  * For each report, three risk components (direct, bidirect, line) are
  * computed as base_risk × temporal decay (logistic in report age), then
  * spread along the report's lines with a spatial decay (beta-binomial in
- * segment distance). Components accumulate additively across reports
+ * segment distance; on circular lines the distance wraps around the ring).
+ * Components accumulate additively across reports
  * (capped at 1.0), are summed into a final risk per segment (capped at
  * 1.0), mapped to a color, and finally propagated to overlapping segments
  * (segments sharing the same unordered station pair).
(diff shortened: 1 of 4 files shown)


johan-t:
@codex review

PR #849, opened 2026-07-20
feat: add historical line insights API

johan-t:
## Why

Line detail needs a stable historical picture rather than reusing the live-report feed, because live reports include predictions and only describe the immediate moment. This endpoint reads observed reports directly and is cacheable for a day, which keeps the UI honest and the request cost predictable.

Sparse line histories can produce misleading spikes. The profile therefore falls back to the citywide profile when a line has fewer than 80 reports for the weekday or less than 24 distinct weeks of coverage. Both gates matter: volume without time coverage can still be seasonal or dominated by a short campaign. Hotspots remain line-specific because replacing station distribution with citywide stations would break the route's geography.

All hour bucketing uses the configured city timezone so the service remains city-agnostic and the displayed commuter rhythm matches local time.

## What changes for users

Clients can request `GET /insights/line/:lineId` and receive:

- a 24-hour weekday rhythm with explicit line-versus-city source metadata;
- all-time, line-specific station shares;
- a stable `LINE_NOT_FOUND` response for unknown lines.

## Validation

- `bun run test -- getInsights.test.ts` — 4 integration tests passed.
- `bun run lint` — passed with three pre-existing comment-capitalization warnings.
- `bun run typecheck` — currently blocked by existing dependency/global-type conflicts; the newly exposed `LINE_NOT_FOUND` union issue was corrected.

@codex review

### DIFF
--- a/packages/api-worker/src/app-env.ts
+++ b/packages/api-worker/src/app-env.ts
@@ -161,7 +161,7 @@ const applyServices = (c: Context<Env>, db: DbConnection, config: AppConfig) => 
 
     c.set('config', config)
     c.set('reportsService', reportsService)
-    c.set('insightsService', new InsightsService(db, transitNetworkDataService))
+    c.set('insightsService', new InsightsService(db, transitNetworkDataService, c.get('city').timezone))
     c.set('riskService', new RiskService(reportsService, transitNetworkDataService))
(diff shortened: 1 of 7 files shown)


johan-t:
@codex review

PR #850, opened 2026-07-20
feat: add line detail experience

johan-t:
## Why

A line view should help riders understand where controls cluster without destroying the route's mental model. Stations are therefore kept in physical line order. Consecutive low-activity stations collapse into expandable groups, preserving distance and sequence while keeping multiple hotspots visible at once.

Provider route variants such as `M1-a` and `M1-b` describe internal routing shapes, not concepts riders should have to understand. The public line name is therefore the identity used by the URL and insights endpoint. The API aggregates reports and stations across every matching variant and returns `variantCount`, making the grouping explicit without exposing variant IDs as public identifiers. The UI uses the longest variant as the route spine and inserts branch-only stations before their next shared stop, so combined data retains a useful geographic order instead of becoming a hotspot ranking.

The chart intentionally shows one typical weekday rather than mixing historical and live bars. A thin current-time marker gives orientation without implying that the historical profile is a live forecast. The nearby report count covers a rolling window, so it is labeled “last 24h” rather than “today.” When line history is too sparse, the heading explicitly names the citywide fallback so confidence is communicated instead of hidden.

The hotspot region scrolls independently and the report action remains fixed. This keeps the primary action reachable on small screens while allowing long routes to remain inspectable. Space is reserved before insights arrive to prevent layout shift, but no skeleton is shown because the modal is compact and the content does not need a simulated preview.

Circular lines omit endpoint direction because a start-to-end label invents a hierarchy that does not exist. Non-circular lines retain endpoint context.

Opening the report flow from a line preselects the public line name. This removes a redundant choice, includes every internal variant, and reduces the risk of reporting against a similarly named route.

## What changes for users

- Lines in station detail open a dedicated, deep-linkable line sheet addressed by public line name.
- Variant reports and stations appear as one combined line instead of separate provider-specific routes.
- The sheet shows a typical local weekday, current-time marker, ordered hotspot clusters, and expandable quieter stations.
- The rolling report count is explicitly labeled as the last 24 hours.
- The report CTA stays visible and opens the report form with the public line selected.
- Line opens, hotspot selections, CTA clicks, and dwell time are tracked for adoption and funnel analysis.

## Validation

- API integration suite: 7 line-insights tests passed, including aggregation of reports created through the public API for every internal variant.
- API lint passed with three pre-existing comment-capitalization warnings.
- Frontend production build passed.
- Frontend lint passed with two existing React Compiler/TanStack Virtual warnings.
- End-to-end local check: `/line/M1` rendered one line from 2 variants with 325 aggregated reports across 41 stations; its CTA opened `/report?lineName=M1` with M1 selected.
- The former variant-ID insights URL returns 404, confirming that the public contract is name-based.

@codex review


johan-t:
@codex review

PR #860, opened 2026-07-21
fix: remove misleading zero-value activity rows

johan-t:
## What changed

Removed the station-wide 24-hour report count and its unused data plumbing while preserving per-line activity. The existing low-share presentation remains truthful: non-zero values below one percent render as `<1%`, and collapsed groups keep their aggregate share visible.

## Why this was necessary

The station-wide count competed visually with the served-line choices and could imply there was no useful activity even when a line had meaningful reports. Separately, rounding a small non-zero hotspot share to `0%` contradicts the hotspot data and makes the ranking appear broken.

## Impact

Station sheets focus on actionable served-line activity, while the narrow line sheet continues to present low-share stations truthfully.

## Validation

- frontend production build
- targeted ESLint and Prettier checks
- in-app browser at 390×844 against production-shaped U7 data: expanded Zwickauer Damm renders as `<1%`
- pairwise three-way merge simulations

Linear: [FRE-760](https://linear.app/freifahren/issue/FRE-760/fix-remove-misleading-zero-value-activity-and-hotspot-rows)

@codex review


johan-t:
@codex review

PR #861, opened 2026-07-21
feat: prefetch served-line insights from station detail

johan-t:
## What changed

When station detail has loaded its served lines, it resolves and deduplicates their public line names and starts background prefetches with the shared `lineInsightsQueryOptions`. TanStack Query owns cache freshness and in-flight deduplication.

## Why this was necessary

Station detail exposes immediate links to every served line, but line insights previously began loading only after navigation. Warming the exact query-cache entries while the user reads the station sheet removes that avoidable wait without blocking rendering or adding a parallel cache mechanism.

## Impact

Line drill-downs from station detail feel immediate; duplicate variants, Strict Mode effect replay, fresh data, and in-flight requests do not create duplicate network work.

## Validation

- frontend production build
- targeted ESLint and Prettier checks
- in-app browser at 390×844: station detail renders immediately and navigates through the full-row U7 link to the warmed line detail
- pairwise three-way merge simulations

Linear: [FRE-756](https://linear.app/freifahren/issue/FRE-756/prefetch-served-line-insights-when-station-detail-opens)

@codex review


johan-t:
@codex review

PR #863, opened 2026-07-22
fix: preserve line detail entry sources

johan-t:
## What changed

- Pass the originating entry point through the line-detail route.
- Record `station`, `report`, and `reports_list` for the currently shipped entry points.
- Reserve `map` and `search` for the planned entry points, and use `direct` for deep links or invalid/missing source values.

## Why this was necessary

`line_detail_opened` was emitted inside `LineDetail` with `source: 'station'` hardcoded, and its TypeScript contract only permitted that value. As a result, opens from report detail and the reports overview were silently misclassified as station traffic, leaving the PostHog source breakdown with only one series.

Keeping capture at the successfully rendered destination preserves the event's meaning while explicit route context provides the missing attribution.

## Validation

- `bun run format:check`
- `bun run lint` (passes with two pre-existing React Compiler warnings)
- `bun run build`
- In-app browser against the local app with live data:
  - station → `/line/U8?source=station`
  - reports overview → `/line/U8?source=reports_list`
  - report detail → `/line/U6?source=report`
  - each route rendered the expected line detail and emitted `line_detail_opened`

No tests were added.

@codex review

### DIFF
--- a/packages/frontend-next/src/components/map/LineDetail.tsx
+++ b/packages/frontend-next/src/components/map/LineDetail.tsx
@@ -8,7 +8,7 @@ import { useStations } from '@/api/transit';
 import { Button } from '@/components/ui/button';
 import { CardContent } from '@/components/ui/card';
 import { useModalViewDuration } from '@/hooks/useModalViewDuration';
-import { track } from '@/lib/analytics';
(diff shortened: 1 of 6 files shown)


johan-t:
@codex review

PR #949, opened 2026-08-16
fix: converge remote D1 on the seeded reference data

johan-t:
## What

The remote D1 load now upserts each reference row on its **natural key** and prunes rows that are no longer in the snapshot, instead of appending with `INSERT OR IGNORE` on the surrogate id.

## Why

The MetroBus lines shipped in #947, but on prod **21 of the 26 bus lines have no segments at all** — the lines exist, the map draws nothing.

It is not the snapshot or the seed logic. Running the real seed against the committed snapshots produces 2147 segments with zero warnings and geometry for all 49 bus relations. Prod has 1574: every rail segment, and only 164 bus ones.

The remote load dumped every row as `INSERT OR IGNORE INTO segments VALUES (...)`, **including the surrogate `id`**. CI rebuilds the local D1 from scratch, so ids are assigned by insertion order — adding 26 bus lines shifted all of them. Prod already had rows at those ids, so `OR IGNORE` silently dropped the new ones. Two independent confirmations:

- The bus segments that survived occupy ids **1411–1655** exactly — the single gap above prod's existing block. Everything above collided and was discarded.
- **1092 ids in prod name a different station pair** than a fresh build produces.

**This is wider than buses.** An additive-only load also means content changes (a moved station, a recoloured line) never propagate, and rows dropped from the network are never removed. Leipzig is currently serving **84 phantom segments** for station pairs its lines no longer have.

## How

- Upsert on the natural key (`segments` on `(line_id, from_station_id, to_station_id)`), and omit the surrogate `id` entirely so a surviving row keeps the id the risk layer uses as its MapLibre feature id.
- Prune via mark-then-sweep — negate the ordering column, let each upsert write the real value back, delete what is still negative — rather than a `NOT IN` holding thousands of tuples in one statement.
- Only the derived tables (`line_stations`, `segments`) are swept. `reports` references `stations` and `lines` with `ON DELETE no action`, so pruning those would either fail the load or orphan user data.
- The dump moves to its own module so it is importable without triggering the seed entrypoint.

## Verification

Reconstructed prod's actual broken state locally (bus segments deleted, ids renumbered, a stale row, a drifted station, a user report) and applied the new dump:

```
identical segment content    : True
bus lines with segments      : 26/26
stale row swept              : True
line_stations converged      : True
stations converged           : True
report preserved             : True
pre-existing ids kept stable : 1410/1411
```

The one id that moves is the stale row, which should be deleted. Plus 7 new unit tests covering the statement shape; full api-worker suite 314 passing.

## Notes

- This only reaches prod on the next deploy, since the reseed runs in the deploy pipeline. The current data stays broken until then.
- Interrupting the load mid-file can leave marker values negative; the next seed re-marks and converges. That is strictly better than today, where an interrupted load leaves rows silently missing.
- The deeper design question — that local and remote seeding take different code paths, which is why local was green while prod was broken — is worth addressing separately rather than in this fix.

### DIFF
--- a/packages/api-worker/src/db/seed/dump-reference-tables.ts
+++ b/packages/api-worker/src/db/seed/dump-reference-tables.ts
@@ -1,0 +1,87 @@
+import type { D1Database } from '@cloudflare/workers-types'
+
+import { toSqlLiteral } from '../sql-literal'
+
+interface ReferenceTableSpec {
+    /** Natural key an existing remote row is matched on. Never a surrogate id — see below. */
+    conflictTarget: readonly string[]
+    /** Columns left out of the statement entirely; the remote row keeps its own value. */
+    omitColumns?: readonly string[]
+    /** Integer column negated to mark rows before the load, so untouched ones can be swept after. */
+    sweepMarkerColumn?: string
+}
+
+/*
+ * How each reference table is reconciled with the freshly built local copy.
+ * Parents before children for FK order.
+ *
+ * `conflictTarget` is deliberately never `segments.id`. That value is assigned by
+ * insertion order into a database CI rebuilds from scratch, so it names a different
+ * segment whenever the network changes. Writing it as a literal and loading with
+ * INSERT OR IGNORE meant any row whose id was already taken remotely was silently
+ * discarded — which is how prod ended up serving 21 MetroBus lines with no geometry.
+ * Matching on the natural key instead also keeps ids stable for rows that persist,
+ * which is what the risk layer uses as its MapLibre feature id.
+ *
+ * `sweepMarkerColumn` marks a table prunable: a row no longer in the snapshot has to
+ * go, or prod keeps drawing a segment for a station pair its line no longer has. Only
+ * the derived tables are swept — `reports` references `stations` and `lines` with
+ * ON DELETE no action, so pruning those could fail the load or orphan user data.
+ */
+export const REFERENCE_TABLES: Record<string, ReferenceTableSpec> = {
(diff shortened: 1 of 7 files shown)


johan-t:
@codex review

johan-t:
Both review points were valid and are fixed in `ae0c981e`.

**Sweep marker was self-inverse.** Correct — `-x - 1` applied twice returns `x`, so a retried load un-marked every row and obsolete ones escaped the sweep. My note about it self-healing on the next seed was wrong. Now a constant `-1`, which is idempotent by construction; every surviving row has its marker rewritten by its own upsert anyway, so nothing is lost by not preserving the value.

**The stand-in D1 never executed the SQL.** Also correct, and it is exactly why the marker bug got through — the assertions only inspected generated text. The suite now runs against the shared D1 from `tests/setup.ts`: it applies the dump to a deliberately drifted database (bus segments deleted, ids renumbered, an obsolete row present) and asserts convergence, id stability and that reports survive. I confirmed the retry test fails against the old marker before restoring the fix.

While verifying, this surfaced a second bug the PR would otherwise have shipped: converging on the natural key **changes segment ids**, and the map joins risk onto the segments GeoJSON by id in the client, where that GeoJSON is persisted to IndexedDB and busted only by the frontend build id. A reseed changes ids without changing the build, so a returning client paints fresh risk colours onto stale segments — visible on this PR's preview, absent on main only because main's load could never change an id. Segment ids are now derived from `line|from|to`, so the same station pair gets the same id in every database and on every rebuild. `tests/riskSegmentIds.test.ts` covers it.

PR #957, opened 2026-08-19
feat: add hamburg as a new city

pmlk:
## Summary

Adds Hamburg as a new city with its transit data.

## What changed

- new database binding `DB_HAMBURG`
- new city config for hamburg
- adds hamburg transit data
    - subways
    - light rails (S-Bahn)
    - metro busses
    - there's no tram in Hamburg

## Notes

- the seed script didn't pick up the following metro bus lines:
    - 1, 2, 3, 13, 15, 21, 22, 32, 40
    - compare with https://www.hvv.de/de/linienfahrplan (query "MB")
- I'm not sure what kind of `promptExamples` would be useful. I'm not aware of any common abbreviations / slang words for stations (like Kotti or Alex in Berlin)

## TODOs

I either don't have access (Cloudflare, Telegram) or am not sure how to do the following things that still need to be done

- [ ] Cloudflare setup
    - [ ] add new subdomain `hamburg.freifahren.org`
    - [ ] add D1 database in Cloudlfare for hamburg
        - [ ] add the `database_id` in `wrangler.jsonc`
- [ ] generate / upload tiles for hamburg
    - [ ] replace the localhost URL in `map.styleUrl` in `hamburg.ts`
- [ ] Telegram setup
    - [ ] create a new telegram channel (`@FreiFahren_HH`)
    - [ ] hook up the existing or a new telegram worker to the new channel
- [ ] Mobile apps? not sure what all needs to be done here


brandesdavid:
Hey Patrick, thanks for the PR.The shape is already right with the own D1, registry, snapshots, MetroBus via the one/two-digit refs. The missing lines (1, 2, 3, 13, …) are probably VHH rather than Hochbahn. I added those operators, but I didn't refresh the snapshots. Will do it after I merge.

What I also put in:

- database_id for api-worker-db-hamburg
- style URL pointing at tiles.freifahren.org instead of localhost
- Hamburg listed: false for now, so Berlin/Leipzig don't see it in the switcher. Local and the PR preview still let you pick it.
- Preview CI: report copy skips if the prod D1 isn't there yet
- Telegram forward skips until a Hamburg chat is mapped
- Frontend route for hamburg.freifahren.org

No need to deploy tiles beforehand. Once this lands on main, the tile job will build Hamburg because packages/cities changed. I added some rough prompt examples; we will tune those once we have real group messages. Telegram channel and the remaining MetroBus lines feel like follow-ups to me, not merge blockers.

thanks

PR #968, opened 2026-08-30
fix: remove unavailable app download prompts

johan-t:
## Summary
- remove the custom iOS App Store download banner
- remove the native Smart App Banner metadata
- remove related translations, analytics events, and dead logic
- keep the warning shown to existing native users that links to the web app

## Validation
- `git diff --check`
- lint/build not run: dependencies are not installed in this worktree (`prettier: command not found`)


johan-t:
Removed since the app was removed from the appstore

PR #970, opened 2026-08-31
fix: stop shipping Telegram extraction prompts in the client bundle

brandesdavid:
## Summary
- The frontend imports the full `CityConfig` object for map bounds/community info/etc. Bundlers can't tree-shake individual properties off an object literal that's otherwise used, so the `telegram` field (Mistral inspector-keyword lists + few-shot prompt examples for Berlin/Hamburg/Leipzig) was shipping verbatim in the public JS bundle — readable by anyone via devtools, independent of repo visibility. Confirmed with a real `vite build` grep before/after.
- Splits `packages/cities` into a frontend-safe half (`*_PUBLIC`, exported from a new `@freifahren/cities/public` entry point) and a telegram-only half (`*.telegram.ts`, backend-only). Root `@freifahren/cities` still exports the full merged `CityConfig` for api-worker/telegram-worker. Updated the 4 frontend-next modules that read city config, plus the tsconfig paths + Vite alias that duplicate the package's exports map for this workspace.
- Small fixes found while finishing the Hamburg Telegram-history eval/import pass:
  - Add `"city"` to the generic-non-station-word list (fixed a fuzzy-match false positive: "Richtung City" resolved to Hamburg's "Rübenkamp (City Nord)").
  - Add an explicit non-report rule for general police/protest/political content in the extraction prompt — the Hamburg source Telegram group is topically broader (police/demos) than a pure fare-control channel.
  - Treat DNS blips (`ENOTFOUND`/`getaddrinfo`/`EAI_AGAIN`) as retryable in the eval harness, not just 429/5xx/timeout.
  - Allow `db:import-telegram` for `hamburg`, not just `leipzig`.

## Test plan
- [x] `bun x tsc --noEmit` clean in `packages/cities`, `packages/telegram-worker`, `packages/api-worker` (no cities/CityConfig-related errors)
- [x] `bun run build` in `packages/frontend-next` succeeds; `grep`ing `dist/` for `Gelbwesten`/`inspectorKeywords`/`promptExamples`/etc. now returns nothing (previously present)
- [x] `packages/telegram-worker`: 61/61 tests pass
- [x] `packages/api-worker`: 257/257 tests pass

### DIFF
--- a/packages/api-worker/src/db/import-telegram-extract.ts
+++ b/packages/api-worker/src/db/import-telegram-extract.ts
@@ -148,8 +148,9 @@ if (cities.length !== 1) {
     throw new Error('--city <slug> is required (one city)')
 }
 const city = cities[0]
-if (city !== 'leipzig') {
-    throw new Error('telegram extract import is Leipzig-only')
+const SUPPORTED_CITIES = new Set(['leipzig', 'hamburg'])
+if (!SUPPORTED_CITIES.has(city)) {
+    throw new Error(`telegram extract import is only supported for: ${[...SUPPORTED_CITIES].join(', ')}`)
 }
 
 importCity(
--- a/packages/cities/package.json
+++ b/packages/cities/package.json
@@ -5,7 +5,8 @@
   "description": "Shared, dependency-free city registry consumed by every FreiFahren toolchain.",
   "type": "module",
   "exports": {
-    ".": "./src/index.ts"
+    ".": "./src/index.ts",
+    "./public": "./src/public.ts"
   },
   "types": "./src/index.ts"
 }
--- a/packages/cities/src/berlin.telegram.ts
+++ b/packages/cities/src/berlin.telegram.ts
@@ -1,0 +1,74 @@
+import type { CityTelegramProfile } from './types'
+
+// Few-shot examples appended to the Telegram extraction prompt. Tuned to teach
+// disambiguation the model gets wrong: slang names, direction-vs-station,
+// all-clear messages, the implicit-direction case. Kept in German since the chat
+// is mixed German/English.
+const promptExamples = `Message: "U2 alex hab in dem bahnstation gesehen"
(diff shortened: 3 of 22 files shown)


johan-t:
Is this a security issue or a bundle size issue?

johan-t:
The changed city configuration file failed the required frontend formatting check, so I applied its Prettier formatting and verified the build, lint, formatting, Telegram tests, and absence of extraction-prompt markers in the client bundle.

brandesdavid:
Primarily security, not bundle size — though we get a modest size win too.

The `telegram` field on each `CityConfig` carries the Mistral extraction prompt: the inspector-keyword lists and few-shot examples used to turn a raw Telegram message into a structured "inspector sighting" report (station, line, direction). Because the frontend imported the full `CityConfig` object for map bounds/community info, and bundlers can't tree-shake individual properties off an object literal that's otherwise used at runtime, that entire field shipped verbatim in the public JS bundle — readable by any visitor via devtools, no repo access needed. Confirmed by grepping a real `vite build` output for `inspectorKeywords`/`promptExamples`/`Gelbwesten` etc. before and after this change.

Why it matters: knowing the exact detection vocabulary makes it easier to either evade extraction (word a real sighting so Mistral doesn't flag it) or manufacture false positives (craft a message that reliably gets extracted as a plausible station+line report, i.e. spam injection into the pipeline). The bundle was a much lower bar for that than reading source — every website visitor gets it "for free," whether or not they'd ever look at GitHub.

This doesn't fully close the loop if the repo itself goes public later — anyone browsing `packages/cities/src/*.telegram.ts` on GitHub would still see the same data, just gated behind repo visibility instead of a browser tab. That's a separate, follow-up decision (e.g. move the prompt data out of the repo into a secret/config store) — flagging it here for visibility, not blocking this PR on it.

brandesdavid:
security by obfuscation is not security - sun tzu

PR #975, opened 2026-09-05
feat: normalize trusted Telegram reports through the API

johan-t:
Telegram reports currently go straight from the Telegram worker to report-gate, bypassing the API's transit post-processing. A report can therefore retain a direction without a usable line (FRE-787). This change makes the API own report normalization for both HTTP and trusted Telegram intake, while keeping report-gate responsible for persistence and the public anti-spam path.

## Design

```text
Public HTTP POST /reports -> ReportSubmissionService -> public report-gate intake
Telegram worker -> API private RPC intake -> ReportSubmissionService -> trusted report-gate intake
```

`ReportSubmissionService` owns the shared normalize → persist → invalidate sequence. The HTTP router gets it from request context. The new `TrustedTelegramReportsEntrypoint` is exported from the API worker with Sentry instrumentation; its RPC adapter builds the equivalent city-specific dependencies because Hono middleware does not run for RPC calls.

The Telegram RPC contract accepts only a city slug and station/line/direction identifiers. It rejects caller-provided source, trust and database descriptors. The API resolves the city/database, assigns `source: telegram`, runs the existing transit rules, and calls `TRUSTED_REPORT_GATE`. Access is through a named Cloudflare service binding, with no public HTTP route or shared password header.

The public schema defaults omitted sources to `web_app` and accepts only `web_app`, `mini_app`, and `mobile_app`. `telegram` is rejected before either gate is called. Missing HTTP city parameters retain the existing default for legacy clients; RPC city is required and unknown cities are rejected.

The gate client validates RPC envelopes and response data, translates unavailable bindings/transport/malformed results into `REPORT_GATE_UNAVAILABLE`, and preserves structured gate rejections. Trusted intake fails closed if its binding is missing, including in preview mode.

Cache invalidation is awaited after persistence and failures do not turn a committed report into an apparent failed submission. RPC targets the canonical API URL and city app origin; other origin/colo variants retain the existing five-minute edge TTL bound.

## Scope and cost

This adds a service-binding hop and reuses the existing API services and D1 tables. It adds no queues, Durable Objects, scheduled jobs, dependencies, or migrations. The companion private-gate PR inserts trusted reports immediately with trust 1 without scoring, identity extraction, Turnstile or Telegram forwarding. Public submissions retain their existing anti-spam behavior.

Bot-token consolidation (FRE-788) and adaptive per-city Telegram summaries are follow-up work. Public-report forwarding remains owned by report-gate in this first change.

## Validation

- API typecheck, full integration suite: 268 tests across 28 files.
- Frontend production build (including TypeScript) passed against the narrowed public input schema; existing large-chunk warnings remain.
- API lint: no errors (warnings remain); full formatting check and Wrangler deployment dry run passed.
- New integration coverage exercises HTTP/RPC normalization parity, direction-without-line regression, public source spoofing, strict RPC validation, missing city database/binding, and cache failure after persistence.
- Ran the actual Telegram, API and private gate bundles together under local Miniflare with real RPC service bindings and D1. Applied API migrations and seeded bundled reference data through `seedBaseData`.
- Sent an authenticated Telegram webhook; mocked only external Mistral/Turnstile responses. Verified normalized null line/direction and immediate trust 1 even with invalid scoring configuration, public read visibility, rejection of public Telegram source, public Turnstile quarantine, unknown-city rejection and no public trusted-intake route.
- Companion suites: Telegram 61 tests and private gate 37 tests, with typechecks and deployment dry runs passing. These are local results; hosted CI and production behavior are separate checks.

## Rollout

Merge/deploy the private gate change first, then this API PR. Confirm the API deployment succeeds before merging the dependent Telegram-worker PR that switches its binding. Separate PRs prevent the independent API and Telegram deployment workflows from racing. No production deployment was performed during this validation.

Rollback order: switch Telegram back before removing the API RPC entrypoint. The existing gate entrypoint and input shape remain compatible with the old Telegram worker.

## Related PRs

1. Private gate: https://github.com/FreiFahren/report-gate/pull/10
2. API ownership: https://github.com/FreiFahren/FreiFahren/pull/975
3. Telegram binding switch (depends on deployed API): https://github.com/FreiFahren/FreiFahren/pull/976

### DIFF
--- a/packages/api-worker/src/app-env.ts
+++ b/packages/api-worker/src/app-env.ts
@@ -6,8 +6,10 @@ import { AppError } from './common/errors'
 import { createLogger, Logger, LogLevel } from './common/logger'
 import { createD1Db, DbConnection } from './db'
 import { InsightsService } from './modules/insights'
-import type { PublicReportGate } from './modules/report-gate/report-gate-contract'
+import type { PublicReportGate, TrustedReportGate } from './modules/report-gate/report-gate-contract'
 import { ReportsService } from './modules/reports'
+import { ReportSubmissionService } from './modules/reports/report-submission-service'
+import { invalidateStationReportsCache } from './modules/reports/reports-cache-middleware'
 import { RiskService } from './modules/risk'
 import type { CacheCtx } from './modules/transit/reference-cache'
 import { TransitNetworkDataService } from './modules/transit/transit-network-data-service'
@@ -17,6 +19,7 @@ export type Bindings = {
     DB?: D1Database
     DB_HAMBURG?: D1Database
     DB_LEIPZIG?: D1Database
+    TRUSTED_REPORT_GATE?: TrustedReportGate
     REPORT_GATE?: PublicReportGate
     REPORT_GATE_MODE?: 'preview-open'
     CORS_ORIGINS?: string
@@ -84,6 +87,7 @@ export type Services = {
 export type Env = {
     Bindings: Bindings
     Variables: Services & {
+        reportSubmissionService: ReportSubmissionService
         logger: Logger
         config: AppConfig
         // This request's city database.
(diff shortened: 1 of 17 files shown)


johan-t:
@codex review

PR #982, opened 2026-09-05
fix: skip repeated location soft asks for returning Safari users

johan-t:
Safari can return `prompt` even when a geolocation request would succeed without a native dialog. Both sharing flows now use one decision: request location when permission is granted or a previous web request succeeded; otherwise preserve the first-time soft ask and explicit-denial handling.

Store only a success flag, clear it on permission denial, and retain it through temporary failures. Native checks remain authoritative. The optional report step continues after an automatic request, matching explicit Allow instead of showing another permission explanation.

Expired browser permissions may still show a native prompt directly. Existing users need one successful request after this change before the fallback is remembered. WebKit background: https://bugs.webkit.org/show_bug.cgi?id=275950#c2

Validation:
- Production build passed (existing large-chunk warning).
- Changed-file ESLint and git diff --check passed.
- `bunx playwright test --config playwright.location.config.ts`: 14 passed across Chromium and WebKit with an iPhone viewport. Tests exercise the real provider and sharing hooks with simulated permission/location APIs, covering reload persistence, first-time asks, granted/unsupported/denied states, denial and timeout.
- Initial cold-server map test timed out; complete Chromium rerun and subsequent two-engine run passed with startup headroom in the delayed-prompt assertion.
- Physical iPhone native permission dialogs have not been tested.

### DIFF
--- a/packages/frontend-next/e2e/fixtures/location.html
+++ b/packages/frontend-next/e2e/fixtures/location.html
@@ -1,0 +1,2 @@
+<div id="root"></div>
+<script type="module" src="./location.tsx"></script>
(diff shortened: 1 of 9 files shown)


johan-t:
@codex review

johan-t (commented):
Reviewed the location permission changes and reran all 14 Chromium/WebKit regression tests successfully; no actionable issues found.

PR #985, opened 2026-09-06
fix: evaluate predicted report thresholds in city timezone

### LINKED ISSUE #901: Predicted-reports threshold uses the process timezone, not the city's
The predicted-reports threshold is evaluated in the **process** timezone rather than the city's. That shows up first as two tests that fail on a developer machine in `Europe/Berlin`, but the same root cause means the feature is running 1–2 hours out of phase in production.

## Reproducing the test failure

```
cd packages/api-worker
TZ=Europe/Berlin bun run test tests/getReports.test.ts   # 2 failed | 17 passed
TZ=UTC           bun run test tests/getReports.test.ts   # 19 passed
```

Failing:

- `Predicted reports threshold > increases threshold from early morning to peak hours`
- `Predicted reports threshold > maintains a linear threshold ramp until the end of each window without premature clamping`

CI runs UTC, so `main` is green there and red for anyone whose machine is set to the timezone of the city we serve.

## Why

The tests pin the instant in UTC:

```ts
const weekdayLateEvening = DateTime.utc(2024, 1, 15, 20, 30)
Settings.now = () => weekdayLateEvening.toMillis()
```

`Settings.now` fixes the *instant*, not the *zone*. The handler then calls `DateTime.now()`, which resolves in luxon's default zone — the process zone — and `calculateBasePredictedReportsThreshold` reads `currentTime.hour` off that.

Under `Europe/Berlin` the intended 20:30 becomes 22:30, which lands in the flat `21:00–07:00 → 1` band instead of on the evening ramp. It then equals the 23:00 case the test compares it against, and `expect(latePredicted).toBeGreaterThan(nightPredicted)` fails with `expected 1 to be greater than 1`.

## The part that is not just a test problem

`reports-routes.ts` passes `DateTime.now()` with the comment *"Intentionally pass in local time"*. The intent is right — the ramp is modelled on commute hours — but Workers run in **UTC**, so "local" in production is UTC, not Berlin.

Concretely, in CEST (UTC+2) the ramp is shifted two hours later than intended:

| Berlin time | intended threshold | actual (UTC-evaluated) |
|---|---|---|
| 07:00–09:00 (morning ramp) | ramps 1 → 7 | 1 (flat minimum) |
| 18:00 (evening decline begins) | 7, starting to fall | 7, no decline yet |
| 21:00–23:00 | 1 (flat minimum) | still on the ramp, ~5 → 1 |

The threshold controls how many **predicted** reports get injected when there are not enough real ones, so the effect is user-visible: historic/synthetic reports are padded into the map at the wrong times of day — most during the late Berlin evening, and least during the morning commute, which is the inverse of what the model intends.

## Suggested fix

`CityConfig` already carries the field for this, and its docstring says exactly what it is for:

```ts
/** IANA timezone used to bucket reports into local service hours. */
timezone: string     // 'Europe/Berlin' for both cities today
```

So the fix is to resolve the current time in the request's city zone rather than the process zone — `DateTime.now().setZone(c.get('city').timezone)` at the two call sites in `reports-routes.ts` — instead of relying on an ambient default. That also makes the tests timezone-independent, since the zone stops depending on where the suite runs, and it removes an assumption about Berlin that AGENTS.md's city-agnostic rule already says belongs in config.

Worth checking whether anything else reads `DateTime.now().hour`/`.weekday` the same way; `post-process-report.ts` uses `DateTime.utc()` explicitly for its station guessing, which is consistent but is a *different* choice from the one this code intends, so the two are worth reconciling deliberately.

<sub>Found while running the suite locally during #900.</sub>

johan-t:
Fixes #901. Resolves FRE-779.

Both reports endpoints now evaluate the predicted-report threshold in the request city's configured timezone. Commute ramps and the weekend adjustment therefore follow city time, including daylight-saving changes, instead of the host process timezone.

Adds nine API integration cases covering winter/summer commute hours, the evening cutoff, the local weekday rollover, and both DST transitions. Each exercises both endpoints for Berlin, Hamburg, and Leipzig under three Luxon default zones. Requests run through routing and real D1, with history created through POST requests; the service spy calls through to the real implementation. All nine cases failed before the fix and pass after it.

Reviewed nearby hour/weekday consumers: insights already uses city time; station guessing compares both historical and requested times in UTC. That separate historical matching behavior remains unchanged.

Validation:
- Full API suite: 277 tests passed under TZ=UTC and 277 under TZ=Europe/Berlin.
- Final regression run after test cleanup: 30 getReports tests passed under TZ=Europe/Berlin.
- API typecheck, formatting, and git diff --check passed. Lint: no errors, 14 existing warnings; changed files lint cleanly.
- Production frontend build and both existing Playwright browser tests passed against local Worker/D1.
- Additional temporary Playwright smoke check passed with REPORT_GATE_MODE=preview-open: POST a report, read real reports through both endpoints, request a future range to exercise historical predictions, and render the report list and station page with no browser JavaScript errors. Screenshot visually inspected.

The existing browser harness omits the report-gate service, so its report reads log 503s; the additional smoke check uses the supported local preview mode to exercise successful report flows. Production report-gate integration and deployment are not part of this validation. No schema, dependency, or frontend changes.

Preview verification (2026-09-06, deployed head `050ee2b5`):
- Preview deployment and all CI checks passed. GitHub deployment 6297515463 points to this exact head.
- Live API checks passed for Berlin, Hamburg, and Leipzig: transit/config, current reports, predicted reports for a future range, and station-scoped predictions. Each prediction response returned the expected nighttime minimum of one report with `isPredicted: true` and `expiresAt: null`.
- Submitted an isolated test report through the deployed browser UI (U8, Alexanderplatz). The success screen appeared and the API returned the persisted real report; report list, current-report summary, and station detail loaded the submission.
- All three cities' report forms loaded their respective transit data. Browser error log was empty.
- First-visit confirmation remained present on map/list/detail pages; backing page content was inspected without accepting that confirmation. Server-clock winter/summer and DST transitions are covered by the local/API regression matrix, not by changing the deployed clock.

Preview: https://frontend-pr-985.freifahren.workers.dev

### DIFF
--- a/packages/api-worker/src/modules/reports/reports-routes.ts
+++ b/packages/api-worker/src/modules/reports/reports-routes.ts
@@ -51,7 +51,7 @@ export const getReports = defineRoute<Env>()({
     handler: async (c) => {
         const reportsService = c.get('reportsService')
         const query = c.req.valid('query')
-        const now = DateTime.now()
+        const now = DateTime.now().setZone(c.get('city').timezone)
 
         c.header('Cache-Control', 'no-store')
 
@@ -63,7 +63,7 @@ export const getReports = defineRoute<Env>()({
                 currentTime: now,
                 viewer: await resolveViewer(c),
             })
-        ) // Intentionally pass in local time
+        )
     },
 })
 
@@ -87,10 +87,10 @@ export const getReportsByStation = defineRoute<Env>()({
                 from: query.from,
                 to: query.to,
                 stationId,
-                currentTime: DateTime.now(),
+                currentTime: DateTime.now().setZone(c.get('city').timezone),
                 viewer: await resolveViewer(c),
             })
-        ) // Intentionally pass in local time
+        )
     },
 })
 
--- a/packages/api-worker/tests/getReports.test.ts
+++ b/packages/api-worker/tests/getReports.test.ts
@@ -8,6 +8,7 @@ import {
     getDefaultReportsRange,
     MAX_REPORTS_TIMEFRAME,
 } from '../src/modules/reports/constants'
+import { calculatePredictedReportsThreshold, ReportsService } from '../src/modules/reports/reports-service'
 import { db, lineStations, lines, reports } from './test-db'
 import { appRequestWithRedirect, sendReportRequest, setSystemTime } from './test-utils'
 
@@ -389,6 +390,7 @@ describe('Predicted reports', () => {
 })
 
 describe('Predicted reports threshold', () => {
+    const originalZone = Settings.defaultZone
     let testStations: string[]
 
     const seedHistoricalData = async () => {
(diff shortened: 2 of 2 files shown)


johan-t:
@codex review

PR #989, opened 2026-09-08
feat: add journey routing with inspection risk per leg

pepperonas:
## Describe the issue:

Answering "should I expect an inspection on the way to X?" currently means reading it off the map: find your line, follow it with your eyes, judge the colours along the way. The data to answer it directly is already here — `predictSegmentRisk` scores every segment, and the graph in `transit/pathfinding.ts` already knows how stations connect. What is missing is the path between two stations and a way to score it.

This is an outside proposal with no prior issue, so the honest first question is whether the direction is wanted at all. A journey view moves the app closer to something people could mistake for a trip planner, which is a product call rather than a technical one. I have tried to make the boundary explicit below, and I am happy to drop this if it does not fit.

`GET /v0/transit/route?from=&to=` returns the legs of a journey with a risk score per segment.

**No second risk model.** `predictSegmentRisk` is unchanged, line for line; its tests pass untouched. Alongside it, `predictRouteRisk` shares the same components, decay tables and overlap propagation, and differs in two ways that follow from what a journey needs:

- **A clock per segment.** The temporal decay runs on the same timescale as a trip — direct risk halves in roughly 17 minutes — so scoring a whole journey at `now` overstates every leg the rider has not reached yet. Each segment is evaluated at its estimated arrival instead.
- **Every requested segment is returned, zeros included.** The map drops green segments because it only paints the rest; a journey has to distinguish "nothing expected" from "could not be determined".

It scores only the requested segments, so it costs O(targets × reports) instead of a pass over the whole network — cheaper than the `/v0/risk` call that already runs per request.

**Routing.** `findPathWithAStar` answers "how many hops apart", which is all `/transit/distance` needs, and stays untouched. `findRoute` sits beside it with a search state of `(station, line)`, which turns a transfer into an ordinary edge with a cost rather than a special case — that is what lets a slower direct line beat a faster one that needs a change. Costs are seconds, so the heuristic is in seconds too (great-circle distance over an upper speed bound) and stays admissible. `MinHeap` is now generic so both searches share it.

**Travel times** are estimates from per-city constants in `@freifahren/cities` (`secondsPerHop` by route type, `transferSeconds`). There is no timetable data in the schema, so this is the only available source of duration, and it is city-specific.

Two decisions worth stating, both commented at the point they apply because they are not visible from the code:

- **Segments are seeded once per line variant in that variant's station order**, so a hop ridden the other way round has no row of its own; both orientations resolve to the same segment. A hop with no segment behind it is reported as *unrated*, never as zero — showing an unscored stretch as safe is the one failure this feature must not have.
- **The overall level is the worst leg, not a combined probability.** One report is spread across neighbouring segments by design, so those values are correlated; assuming independence would read ten legs at 0.3 as 0.97 for a single inspector.

Not cached at the edge: the answer depends on the viewer's trust threshold and on the current time, like `/v0/risk`. A test pins the `no-store` header so it cannot be added to the cacheable transit paths by accident.

Identical origin and destination is rejected in the query schema rather than through a new internal error code — `INTERNAL_CODES` is a shared closed list and this edge case does not warrant a permanent entry in it.

**Deliberately out of scope:** no alternative-route comparison (the response shape does not preclude adding it), no address or POI search, no walking transfers between nearby stations with different names, and explicitly not a journey planner — durations are estimates and the UI says so.

**Testing.** 23 new tests, 301 total, all green; nothing existing was modified. Unit tests cover the two pure functions (there is precedent in `risk-model.test.ts` and `build-variants.test.ts`, and transfer-cost edge cases are not reachable over HTTP); the behaviour lives in `getRoute.test.ts`, driven through the API with fixtures found in the seeded data rather than named, so it works for any city.

Two things the tests learned from running against real data rather than mocks:

- The timing test builds its own isolated line. On the seeded network, two journeys over "the same" stretch legitimately come back on **different lines** — a tram alternative won between two stations of the intended line — sharing no segments and making the comparison vacuous.
- The report in that test sits at the far end of the line so the shared segments stay below the model's cap. At a saturated 1.0, no amount of elapsed time can show a difference.

**An honest measurement about the per-segment clock.** It works, but it matters less than it sounds. Measured against real Berlin data with a report at the destination, comparing the same final segment reached from different starting points:

| travel time to that segment | risk | vs. shortest |
|---:|---:|---:|
| 2 min | 0.438 | — |
| 10 min | 0.433 | −0.005 |
| 22 min | 0.414 | −0.024 |
| 43 min | 0.301 | −0.137 (−31%) |

On a typical city journey the difference is far below the resolution of the risk bands and invisible to users; it only becomes substantial on the longest trips. That is the correct shape — the decay curve is deliberately flat in the first minutes, and on a short journey almost nothing *should* change — but it is worth saying plainly rather than selling the feature on it.

**Verified live** against a local D1 with the full Berlin seed: a report mid-route produces a clean curve peaking at the reported station (0.438 → 0.731 → 0.581), and unrelated routes stay at exactly 0.000 — including one that ends at the same station as the reported line but uses different segments.

The frontend for this lives on a separate branch and will follow as its own PR once this one is settled, so this can be reviewed on its own.

---

https://claude.ai/code/session_01A6ymvG1kij8CSg49oQyNwp

### DIFF
--- a/packages/api-worker/README.md
+++ b/packages/api-worker/README.md
@@ -74,6 +74,15 @@ development, run all three Worker configs together; only this API is exposed:
 bun run dev:with-report-gate
 ```
 
+The `report-gate` Worker lives outside this repository, so that command only works with access to
+it. Without it, every route that resolves a viewer — `/reports`, `/risk`, `/transit/route` — answers
+`503 REPORT_GATE_UNAVAILABLE`. To work on those paths locally, put the switch that PR previews use
+into `.dev.vars` and run the plain `bun run dev`:
+
+```sh
+echo 'REPORT_GATE_MODE=preview-open' >> .dev.vars
+```
+
 The Stripe webhook signing secret remains an API Worker secret:
 
 ```sh
--- a/packages/api-worker/src/app-env.ts
+++ b/packages/api-worker/src/app-env.ts
@@ -12,6 +12,7 @@ import { ReportSubmissionService } from './modules/reports/report-submission-ser
 import { invalidateStationReportsCache } from './modules/reports/reports-cache-middleware'
 import { RiskService } from './modules/risk'
 import type { CacheCtx } from './modules/transit/reference-cache'
+import { RouteService } from './modules/transit/route-service'
 import { TransitNetworkDataService } from './modules/transit/transit-network-data-service'
 
 export type Bindings = {
@@ -82,6 +83,7 @@ export type Services = {
     insightsService: InsightsService
     riskService: RiskService
     transitNetworkDataService: TransitNetworkDataService
+    routeService: RouteService
 }
 
 export type Env = {
@@ -182,6 +184,7 @@ export const createCityServices = (db: DbConnection, city: CityConfig, cacheCtx:
         reportsService,
         insightsService: new InsightsService(db, transitNetworkDataService, city.timezone),
         riskService: new RiskService(reportsService, transitNetworkDataService),
+        routeService: new RouteService(transitNetworkDataService, reportsService, city.routing),
     }
 }
 
@@ -211,6 +214,7 @@ const applyServices = (c: Context<Env>, db: DbConnection, config: AppConfig) => 
     c.set('insightsService', services.insightsService)
     c.set('riskService', services.riskService)
     c.set('transitNetworkDataService', services.transitNetworkDataService)
+    c.set('routeService', services.routeService)
 }
 
 export const registerContext = (app: Hono<Env>) => {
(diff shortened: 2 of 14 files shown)


johan-t:
Hey, we really appreciate the support and your contribution!

We’ve actually thought about this feature before and used to have it implemented. What we noticed, however, was that usage was quite low. The main reason is that Google Maps and Apple Maps already provide such a good routing experience that it’s difficult for us to offer something at a comparable level, which is ultimately the standard we’d need to meet for people to actually use the feature.

Reaching and maintaining that level of quality would require a significant amount of development and maintenance effort, which we unfortunately don’t have the capacity for at the moment. :(

That said, I really like the idea, and if you notice any other issues with the app, please feel free to report them — or, of course, fix them yourself and contribute :)

PR #990, opened 2026-09-08
feat: check inspection risk for a planned journey

pepperonas:
## Describe the issue:

Follow-up to #989, which added `GET /v0/transit/route`. This puts a destination field in front of it: enter where you are going and see whether inspections are likely on the way, leg by leg.

This branch is cut from `main` and touches only `packages/frontend-next` (plus one CI step and a README line) — no backend files, and nothing here imports from `api-worker`, so it reviews on its own. It does need #989 to return data; without that endpoint the view has nothing to show.

A journey view at `/route/$fromId/$toId`, following the pattern of `/station/$stationId` — the map stays behind it, the loader validates both ids against the station list, and the URL is shareable.

**The view renders and computes nothing.** No risk model in the bundle, and no way for a journey to disagree with the map underneath it. Levels come from the same `riskLevel` the map legend uses, so both speak one language.

**Entry point.** One field for the destination; the start is pre-filled from the nearest station via the existing geolocation path and `closestStationId`, and stays editable — which is also the path when location is unavailable or denied.

**Uneventful stops collapse.** A long journey is mostly quiet, and a screen full of "nothing reported" buries the two stops that matter. Consecutive quiet stops fold into one summary row that expands on tap: an 18-stop journey went from 18 rows and a 748px card to 3 rows and 394px. Two rules keep the summary from lying, both pinned by tests: a stop that could not be assessed is never folded into "nothing reported", and a lone quiet stop stays expanded because collapsing one row saves nothing and costs a tap.

**Honesty in the display**, matching the endpoint's contract: a hop the API could not score reads "no data" rather than risk-free, the duration always carries "approx.", and a footnote states that travel times are estimates and not a timetable.

**The map highlight** goes through a small module store. The result view lives in the map layout's outlet while the layer that paints it lives inside the lazily loaded map, so the two cannot pass props — the same reason `useRiskLayer` keeps its state that way. Segment geometry is already in the client cache, so the layer filters what is there rather than fetching anything.

Two small refactors, both to make something testable or reusable rather than for their own sake: `matchStations` moved out of the search component into `lib/` (it could not be exported from a component file without tripping the fast-refresh rule), and the search's result list became a shared `StationPicker` that both the map search and the planner use.

**This package had no unit test setup** — two Playwright specs, both about the PWA shell — so 13k lines of app logic rested on types and lint alone. This adds Vitest for `src/` only, browserless, ~350ms. `bun run test` still means Playwright; the new step is `bun run test:unit` and runs in CI before the build, so a broken invariant fails in seconds rather than after a bundle and a browser download.

32 tests cover the pieces that would break silently: the search's result cap and empty-query behaviour, the collapse rules above, the highlight store's no-op on unchanged ids (the journey refetches on a timer and would otherwise re-render the map layer every 30s) and its stable snapshot reference, the query key differing per direction, and the risk thresholds shared with the map. Every mutation tried against them fails as intended.

**Two bugs found by measuring rather than reading**, both fixed here:

- The planner button stayed mounted underneath the result card and was unreachable there — the backdrop covers its upper half, the card its lower one. At 1200px this went unnoticed because the card starts lower; at 390px they overlap directly.
- The fade that marks a list continuing below the fold was keyed on the journey data, so expanding a collapsed run changed the height without triggering it — exactly the case that creates the overflow. It now watches the rendered legs through a `ResizeObserver`.

**Verified** at 390px and 1200px: no horizontal overflow, the card fits, Escape/backdrop/close button all dismiss it, and a full pass through search → result → close → reopen works. The map highlight itself was not verified automatically — the MapLibre instance is not reachable from the test harness — only that it produces no errors.

No new runtime dependency; Vitest is dev-only.

---

https://claude.ai/code/session_01A6ymvG1kij8CSg49oQyNwp


johan-t:
closed via this comment: https://github.com/FreiFahren/FreiFahren/pull/989#issuecomment-5653235757

